PR #23, opened 2026-02-03
Migrate Docker registry handler to AWS SDK v2

JamieMagee:
Replaces aws-sdk-go (v1) with aws-sdk-go-v2 for ECR authentication in the Docker registry handler.

## What changed

- Swapped `session.NewSession()` for `config.LoadDefaultConfig()`
- Static credentials now use `credentials.NewStaticCredentialsProvider()`
- `GetAuthorizationToken` takes a `context.Context` parameter
- Defined an `ECRClient` interface since v2 removed the `ecriface` package
- Pointer dereferencing uses `aws.ToString()` helper
- `AuthorizationData` is now a slice of values, not pointers

## Why

AWS deprecated v1 in July 2024 and recommends migrating to v2.


jeffwidman (commented):
👍 Makes sense to me other than the Copilot suggestion. It's been a minute since I worked with go contexts, that what it's suggesting sounds plausible at least. Tag me for an updated review once you address that.

jeffwidman:
We've got a false positive security alert https://github.com/dependabot/proxy/security/code-scanning/23 (requires elevated permissions to view) which will go away when this is merged. The alert is about AWS vulnerability in the S3 v1 SDK go library, but we only use the ECR stuff, so we're not actually vulnerable... but regardless, it should simply go away when this is merged and bumps us to V2

JamieMagee:
Superseded by https://github.com/dependabot/proxy/pull/195

PR #27, opened 2026-02-04
fix: public github release downloads should not be authenticated

### LINKED ISSUE #15: Investigate proxy issue preventing gradle wrapper url validation
### Code improvement description

We currently disable Gradle Wrapper distribution URL validation (`--no-validate-url`) as a workaround because the wrapper task’s URL validation fails when running behind Dependabot’s proxy.

https://github.com/dependabot/dependabot-core/blob/e7b8811f3496ae5e4d741dd246aab721884228b4/gradle/lib/dependabot/gradle/file_updater/wrapper_updater.rb#L126

**Context / Background**
When the Gradle `wrapper` task runs, Gradle validates the configured distribution URL by issuing a `HEAD` request, for example:

`HEAD https://services.gradle.org/distributions/gradle-9.3.0-bin.zip`

This head request then redirects to GitHub and ultimately fails with 

> WARN: Cannot read TLS response from mitm'd server tls: first record does not look like a TLS handshake

Logs 

> proxy | 2026/01/28 13:48:55 [025] HEAD https://services.gradle.org:443/distributions/gradle-9.3.0-bin.zip
>   proxy | 2026/01/28 13:48:55 [025] 307 https://services.gradle.org:443/distributions/gradle-9.3.0-bin.zip
>   proxy | 2026/01/28 13:48:55 [027] HEAD https://github.com:443/gradle/gradle-distributions/releases/download/v9.3.0/gradle-9.3.0-bin.zip
>   proxy | 2026/01/28 13:48:55 [027] * authenticating git server request (host: github.com)
>   proxy | 2026/01/28 13:48:55 [027] 302 https://github.com:443/gradle/gradle-distributions/releases/download/v9.3.0/gradle-9.3.0-bin.zip
>   proxy | 2026/01/28 13:48:55 [029] HEAD https://objects.githubusercontent.com:443/github-production-release-asset-2e65be/696192900/b5669ad6-764c-4b3a-a4eb-d6e13f57ec47?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=releaseassetproduction%2F20260128%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20260128T134856Z&X-Amz-Expires=1800&X-Amz-Signature=63661c31353f570bf663e5fcab68ebc77bb23b0b7ff5c9f390ef8b559a1d23ac&X-Amz-SignedHeaders=host&response-content-disposition=attachment%3B%20filename%3Dgradle-9.3.0-bin.zip&response-content-type=application%2Foctet-stream
>   proxy | 2026/01/28 13:48:55 [029] WARN: Cannot read TLS response from mitm'd server tls: first record does not look like a TLS handshake
>   proxy | 2026/01/28 13:48:55 [031] HEAD https://objects.githubusercontent.com:443/github-production-release-asset-2e65be/696192900/b5669ad6-764c-4b3a-a4eb-d6e13f57ec47?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=releaseassetproduction%2F20260128%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20260128T134856Z&X-Amz-Expires=1800&X-Amz-Signature=63661c31353f570bf663e5fcab68ebc77bb23b0b7ff5c9f390ef8b559a1d23ac&X-Amz-SignedHeaders=host&response-content-disposition=attachment%3B%20filename%3Dgradle-9.3.0-bin.zip&response-content-type=application%2Foctet-stream
>   proxy | 2026/01/28 13:48:55 [031] WARN: Cannot read TLS response from mitm'd server tls: first record does not look like a TLS handshake

To keep updates working, we pass `--no-validate-url` and set `validateDistributionUrl=false`, which skips the distribution URL validation step.

This workaround only seems necessary due to a proxy-related issue. When running without the Dependabot proxy (for example, through my own proxy), the validation works as expected.

yeikel:
The proxy should not assume that asset release downloads require authentication. These endpoints are public, and including a token can actually cause access to fail with an unauthorized error.

For additional context and analysis see https://github.com/dependabot/proxy/issues/15#issuecomment-3839157204

Fixes https://github.com/dependabot/proxy/issues/15

Logs: 

### Before this change

```
2026/02/04 06:50:39 proxy starting, commit: dev
2026/02/04 01:50:39 Listening (:1080)
2026/02/04 01:51:02 [002] HEAD https://services.gradle.org:443/distributions/gradle-9.2.0-bin.zip
2026/02/04 01:51:02 [002] 307 https://services.gradle.org:443/distributions/gradle-9.2.0-bin.zip
2026/02/04 01:51:02 [004] HEAD https://github.com:443/gradle/gradle-distributions/releases/download/v9.2.0/gradle-9.2.0-bin.zip
2026/02/04 01:51:02 [004] * authenticating git server request (host: github.com)
2026/02/04 01:51:02 [004] 302 https://github.com:443/gradle/gradle-distributions/releases/download/v9.2.0/gradle-9.2.0-bin.zip
2026/02/04 01:51:03 [006] HEAD https://objects.githubusercontent.com:443/github-production-release-asset-2e65be/696192900/59c24fc6-8617-4334-917a-58d539c557d3?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=releaseassetproduction%2F20260204%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20260204T065104Z&X-Amz-Expires=1800&X-Amz-Signature=f20b2f55b452f1b7ca94086a2c62f17bcba2b0ee382e77d322fdbd50a243e527&X-Amz-SignedHeaders=host&response-content-disposition=attachment%3B%20filename%3Dgradle-9.2.0-bin.zip&response-content-type=application%2Foctet-stream
2026/02/04 01:51:03 [006] 401 https://objects.githubusercontent.com:443/github-production-release-asset-2e65be/696192900/59c24fc6-8617-4334-917a-58d539c557d3?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=releaseassetproduction%2F20260204%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20260204T065104Z&X-Amz-Expires=1800&X-Amz-Signature=f20b2f55b452f1b7ca94086a2c62f17bcba2b0ee382e77d322fdbd50a243e527&X-Amz-SignedHeaders=host&response-content-disposition=attachment%3B%20filename%3Dgradle-9.2.0-bin.zip&response-content-type=application%2Foctet-stream
```

### After this change

```
2026/02/04 06:49:02 proxy starting, commit: dev
2026/02/04 01:49:02 Listening (:1080)
2026/02/04 01:49:35 [002] HEAD https://services.gradle.org:443/distributions/gradle-9.2.0-bin.zip
2026/02/04 01:49:35 [002] 307 https://services.gradle.org:443/distributions/gradle-9.2.0-bin.zip
2026/02/04 01:49:35 [004] HEAD https://github.com:443/gradle/gradle-distributions/releases/download/v9.2.0/gradle-9.2.0-bin.zip
2026/02/04 01:49:35 [004] 302 https://github.com:443/gradle/gradle-distributions/releases/download/v9.2.0/gradle-9.2.0-bin.zip
2026/02/04 01:49:35 [006] HEAD https://release-assets.githubusercontent.com:443/github-production-release-asset/696192900/59c24fc6-8617-4334-917a-58d539c557d3?sp=r&sv=2018-11-09&sr=b&spr=https&se=2026-02-04T07%3A30%3A36Z&rscd=attachment%3B+filename%3Dgradle-9.2.0-bin.zip&rsct=application%2Foctet-stream&skoid=96c2d410-5711-43a1-aedd-ab1947aa7ab0&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skt=2026-02-04T06%3A30%3A32Z&ske=2026-02-04T07%3A30%3A36Z&sks=b&skv=2018-11-09&sig=kasSevSu%2BWrZDWMK1tepgTDuLhkDuakwTTnEBX2F4f4%3D&jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmVsZWFzZS1hc3NldHMuZ2l0aHVidXNlcmNvbnRlbnQuY29tIiwia2V5Ijoia2V5MSIsImV4cCI6MTc3MDE5MTM3NywibmJmIjoxNzcwMTg3Nzc3LCJwYXRoIjoicmVsZWFzZWFzc2V0cHJvZHVjdGlvbi5ibG9iLmNvcmUud2luZG93cy5uZXQifQ.3DIcTnpywlfoHQyUxPWmo9214Qe8zClbdFZTxtSNDmk&response-content-disposition=attachment%3B%20filename%3Dgradle-9.2.0-bin.zip&response-content-type=application%2Foctet-stream
2026/02/04 01:49:35 [006] 200 https://release-assets.githubusercontent.com:443/github-production-release-asset/696192900/59c24fc6-8617-4334-917a-58d539c557d3?sp=r&sv=2018-11-09&sr=b&spr=https&se=2026-02-04T07%3A30%3A36Z&rscd=attachment%3B+filename%3Dgradle-9.2.0-bin.zip&rsct=application%2Foctet-stream&skoid=96c2d410-5711-43a1-aedd-ab1947aa7ab0&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skt=2026-02-04T06%3A30%3A32Z&ske=2026-02-04T07%3A30%3A36Z&sks=b&skv=2018-11-09&sig=kasSevSu%2BWrZDWMK1tepgTDuLhkDuakwTTnEBX2F4f4%3D&jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmVsZWFzZS1hc3NldHMuZ2l0aHVidXNlcmNvbnRlbnQuY29tIiwia2V5Ijoia2V5MSIsImV4cCI6MTc3MDE5MTM3NywibmJmIjoxNzcwMTg3Nzc3LCJwYXRoIjoicmVsZWFzZWFzc2V0cHJvZHVjdGlvbi5ibG9iLmNvcmUud2luZG93cy5uZXQifQ.3DIcTnpywlfoHQyUxPWmo9214Qe8zClbdFZTxtSNDmk&response-content-disposition=attachment%3B%20filename%3Dgradle-9.2.0-bin.zip&response-content-type=application%2Foctet-stream
```

### DIFF
--- a/internal/handlers/git_server.go
+++ b/internal/handlers/git_server.go
@@ -333,6 +333,10 @@ func getCredentialsForRequest(r *http.Request, credentials *gitCredentialsMap, e
 		return nil
 	}
 
+	if isPublicGitHubDownload(host, r.URL.Path) {
+		return nil
+	}
+
 	// Get credentials for the host that not unscoped to specific repositories.
 	hostCreds := credentials.get(host)
 	credsForRequest := hostCreds.getCredentialsForRepo(allReposScopeIdentifier)
@@ -351,6 +355,12 @@ func getCredentialsForRequest(r *http.Request, credentials *gitCredentialsMap, e
 	}
 
 	return credsForRequest
+}
+
+// GitHub release download URLs are public
+// and do not require authentication
+func isPublicGitHubDownload(host string, path string) bool {
+	return host == "github.com" && strings.Contains(path, "/releases/download/")
 }
 
 // HandleResponse handles retrying failed auth responses with alternate credentials
--- a/internal/handlers/git_server_test.go
+++ b/internal/handlers/git_server_test.go
@@ -127,6 +127,29 @@ func TestGitServerHandler(t *testing.T) {
 		"valid github request")
 }
 
+func TestGitServerPublicReleaseDownload(t *testing.T) {
+	installationCred := testGitSourceCred("github.com", "x-access-token", "v1.token")
+	gheCred := testGitSourceCred("ghe.some-corp.com", "x-access-token", "corp")
+
+	credentials := config.Credentials{
+		installationCred,
+		gheCred,
+	}
+	handler := NewGitServerHandler(credentials, nil)
+
+	req := httptest.NewRequest("HEAD", "https://github.com/gradle/gradle-distributions/releases/download/v9.3.0/gradle-9.3.0-bin.zip", nil)
+	req, _ = handler.HandleRequest(req, nil)
+	assertUnauthenticated(t, req, "Public github.com release downloads should not be authenticated")
+
+	req = httptest.NewRequest("HEAD", "https://ghe.some-corp.com/gradle/gradle-distributions/releases/download/v9.3.0/gradle-9.3.0-bin.zip", nil)
+	req, _ = handler.HandleRequest(req, nil)
+	assertHasBasicAuth(t, req,
+		gheCred.GetString("username"),
+		gheCred.GetString("password"),
+		"valid github request")
+
+}
+
 func TestGitServerHandler_AuthenticatedAccessToGitHubRepos(t *testing.T) {
 	installationToken1 := "v1.token1"
 	privateRepo1Cred := testGitSourceCred("github.com", "x-access-token", installationToken1, withAccessibleRepos([]string{"github/private-repo-1"}))


internal/handlers/git_server.go
```diff
@@ -343,6 +349,10 @@ func getCredentialsForRequest(r *http.Request, credentials *gitCredentialsMap, e
 	return credsForRequest
 }
 
+func isGitHubReleaseDownload(path string) bool {
+	return strings.Contains(path, "/releases/download/")
+}
+
```
yeikel: This approach doesn’t feel entirely safe, as it could also match URLs like `myExample.com/releases/download/`. 

However, I didn’t see any reliable way to verify that the host is GitHub-related. 

For example, `github.com` is obvious, but GitHub Enterprise and other environments are less straightforward

internal/handlers/git_server.go
```diff
@@ -327,6 +327,12 @@ func getCredentialsForRequest(r *http.Request, credentials *gitCredentialsMap, e
 	hostCreds := credentials.get(host)
 	credsForRequest := hostCreds.getCredentialsForRepo(allReposScopeIdentifier)
 
+	// GitHub release download URLs are public
+	// and do not require authentication
+	if len(credsForRequest) != 0 && isGitHubReleaseDownload(r.URL.Path) {
```
yeikel: My attempt at preventing potential cross-type detection is to ensure that this is considered a git host and that we have stored credentials for it. I don't know if this is good enough

internal/handlers/git_server.go
yeikel: @JamieMagee I'd appreciate your input here and in the pull request altogether

internal/handlers/git_server.go
JamieMagee: This worries me for private repos. If someone has a private GitHub repo with release assets, they need auth on the initial request to `github.com` to even get the 302 redirect URL. This returns `nil` for all release downloads, public or private.

The existing retry logic in `HandleResponse` handles 401/403/404 by stripping auth and retrying, but that only applies to the same request. The one failing here is the redirect target (`objects.githubusercontent.com`), which is a different request from the proxy's perspective.

Have you tested this against a private repo with release assets?

internal/handlers/git_server.go
JamieMagee: I agree. A few possible options I can think of:

- Scope this to only github.com (not a complete fix since GHES exists, but narrows the blast radius)
- Instead of skipping auth on the initial request, handle the auth failure on the redirect target. Though that's trickier since it's a separate request through the proxy
- Strip the Authorization header on redirect responses from known GitHub hosts (302 -> asset CDN) rather than preventing auth on the original request

yeikel: > Strip the Authorization header on redirect responses from known GitHub hosts (302 -> asset CDN) rather than preventing auth on the original request

Unfortunately, this does not work because the redirect GitHub gives depends on whether you pass authentication or not. It is an interesting behavior/I ignore why that is: 

> HEAD request without auth: https://github.com:443/gradle/gradle-distributions/releases/download/v9.3.0/gradle-9.3.0-bin.zip → Redirects to https://release-assets.githubusercontent.com/github-production-release-asset..., which resolves correctly.
> 
> HEAD request with auth(it does not matter if it is valid): Same URL → Redirects to https://objects.githubusercontent.com/…, which doesn’t resolve and consistently returns 401.

internal/handlers/git_server.go
yeikel: > Have you tested this against a private repo with release assets?

I tested this and it does not work in its current form. However, the behavior is consistent with GitHub’s documented expectations.

Artifacts from private repositories cannot be downloaded via "public" (what we can see from the browser) release asset URL (for example, `https://github.com/gradle/gradle-distributions/releases/download/v9.3.0/gradle-9.3.0-bin.zip`), even when authentication is provided. GitHub instead requires private release assets to be accessed through the GitHub API and downloaded using the `ASSET-ID`. In the proxy, the closest equivalent would be the `github_api` handler.

I may be mistaken, but from the proxy’s perspective this appears to be out of scope. Rather, the client (such as Dependabot Core) would need to be aware of this in advance and issue an API call instead. 

See: https://github.com/orgs/community/discussions/47453

internal/handlers/git_server.go
yeikel: I tested this on GitHub Enterprise, and the behavior is slightly different. In our environment, SAML/SSO is required, so public download URLs do not work outside of a browser session. I can only assume that even there, the download must be done using the GitHub API

yeikel: > Scope this to only github.com (not a complete fix since GHES exists, but narrows the blast radius)

I updated it to scope it only for github.com. But as described above, in GHE this won't work because no client in GHE should try to reach public urls (at least not if using SSO). That said, I am not sure about all the combinations or deployment options

jeffwidman: I think scoping to only github.com is fine for now, it's definitely a step forwards from the current state and doesn't prevent improving it in the future.

jeffwidman:
👋 I'm back from sabbatical.

This problem seems legit to me... I know I've been bit/annoyed by it before.

What is needed to move this forward?

PR #28, opened 2026-02-07
Add support for bearer token auth with Composer registries

hype09:
This PR follows from the discussion at https://github.com/dependabot/dependabot-core/pull/12946.

Some composer registries, ex. `packages.shopware.com`, only support Bearer authentication. Dependabot currently only supports Basic Auth.

I have added a `token` field to the credentials which takes precedence over basic auth when a token is set and not empty.

The added tests all run on my machine, and I have tested my changes using `dependabot/cli` with the following configuration:

```yml
# config.yml
job:
    package-manager: composer
    allowed-updates:
      - update-type: all
    source:
        provider: github
        repo: <my-repo>
        directory: /
credentials:
  - type: git_source
    host: github.com
    username: x-access-token
    password: $GITHUB_TOKEN

  - type: composer_repository
    registry: packages.shopware.com
    token: $SHOPWARE_TOKEN
```

I ran it with a locally built image of the proxy as follows:

```shell
#!/bin/bash
dependabot update \
    --proxy-image dependabot-proxy-local:dev \
    -f config.yml
```

The log output shows that authentication has worked for `packages.shopware.com`:

```
proxy | 2026/02/07 08:24:03 [012] GET https://packages.shopware.com:443/packages.json
proxy | 2026/02/07 08:24:03 [012] * authenticating composer registry request (host: packages.shopware.com)
proxy | 2026/02/07 08:24:06 [012] 200 https://packages.shopware.com:443/packages.json
```

Disclaimer: My experience with Go is somewhat limited, so if there are more idiomatic ways to solve things, I'd be happy to learn! Also, I've seen a few places in `dependabot-core` where Composer uses basic auth. I'm unsure if we should add token support there, too? If that's the case, I'd be happy so work on a PR.

### DIFF
--- a/internal/handlers/composer.go
+++ b/internal/handlers/composer.go
@@ -93,10 +93,11 @@ func (h *ComposerHandler) HandleRequest(req *http.Request, ctx *goproxy.ProxyCtx
 			continue
 		}
 
-		logging.RequestLogf(ctx, "* authenticating composer registry request (host: %s)", req.URL.Hostname())
 		if cred.token != "" {
+			logging.RequestLogf(ctx, "* authenticating composer registry request (host: %s, token auth)", req.URL.Hostname())
 			req.Header.Set("Authorization", "Bearer "+cred.token)
 		} else {
+			logging.RequestLogf(ctx, "* authenticating composer registry request (host: %s, basic auth)", req.URL.Hostname())
 			req.SetBasicAuth(cred.username, cred.password)
 		}
 


ahtizazhasnain-hue (approved):
M

internal/handlers/composer.go
```diff
@@ -92,7 +94,11 @@ func (h *ComposerHandler) HandleRequest(req *http.Request, ctx *goproxy.ProxyCtx
 		}
 
 		logging.RequestLogf(ctx, "* authenticating composer registry request (host: %s)", req.URL.Hostname())
```
JamieMagee: The log message doesn't distinguish between bearer and basic auth. The NPM handler does this `(host: %s, basic auth)` vs `(host: %s, token auth)`, and it's genuinely useful when debugging why a registry is rejecting credentials. Could you split the log line so it says which auth method is being used?

hype09: Makes sense to me 👍

PR #45, opened 2026-02-10
Bump github.com/elazarl/goproxy to v1.7.2

JamieMagee:
Upgrades `github.com/elazarl/goproxy` from `v0.0.0-20240726154733-8b0c20506380` to `v1.7.2`.

The Dependabot PR in #14 tries to go straight to v1.8.1, but that version has a build failure (see [elazarl/goproxy#745](https://github.com/elazarl/goproxy/issues/745)). The fix landed upstream in [elazarl/goproxy#749](https://github.com/elazarl/goproxy/pull/749) but hasn't been released yet.

v1.7.2 is the latest version that works.


kbukum1:
@JamieMagee , do we need this Bump now? I did merge the following one. Let me know if there is issue.

- https://github.com/dependabot/proxy/pull/76

JamieMagee:
@kbukum1 nope! This was done to upgrade us to as new a version as possible, while avoiding a bug in the newest version. Looks like it was fixed.

PR #47, opened 2026-02-12
Implements Cloudsmith OIDC

joniumGit:
**Implements support for OIDC authentication to the private registry service [Cloudsmith](https://cloudsmith.com/)**

The implementation is based on the information available in the [official documentation](https://docs.cloudsmith.com/authentication/openid-connect#token-exchange) and in the [vendor provided GitHub Action](https://github.com/cloudsmith-io/cloudsmith-cli-action/blob/master/src/oidc-auth.js#L215). The code is mostly copied from the existing JFrog OIDC implementation and adapted to work with Cloudsmith as I am not a Go developer and not that familiar with it.

**Notes:**
* I added an environment lookup to the credential creation to have a fallback value for the audience based on the [GHA implementation](https://github.com/cloudsmith-io/cloudsmith-cli-action/blob/master/src/main.js#L18). I am not sure if this is the right place for it or should it go into the `actions_oidc` code.
* I named input parameters based on the vendor GHA and code variables based on the API parameters.
* I have not been able to test the implementation in practice as I don't know how to run Dependabot locally. 

I appreciate and welcome all feedback on the PR.

### DIFF
--- a/internal/oidc/actions_oidc.go
+++ b/internal/oidc/actions_oidc.go
@@ -614,7 +614,7 @@ func GetCloudsmithAccessToken(ctx context.Context, params CloudsmithOIDCParamete
 		return nil, fmt.Errorf("failed to read cloudsmith token response body: %w", err)
 	}
 
-	if resp.StatusCode != http.StatusOK {
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusCreated {
 		return nil, fmt.Errorf("cloudsmith returned status %d: %s", resp.StatusCode, string(body))
 	}
 
--- a/internal/oidc/actions_oidc.go
+++ b/internal/oidc/actions_oidc.go
@@ -618,7 +618,7 @@ func GetCloudsmithAccessToken(ctx context.Context, params CloudsmithOIDCParamete
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("Cloudsmith returned status %d: %s", resp.StatusCode, string(body))
+		return nil, fmt.Errorf("cloudsmith returned status %d: %s", resp.StatusCode, string(body))
 	}
 
 	var tokenResp cloudsmithTokenResponse
@@ -627,7 +627,7 @@ func GetCloudsmithAccessToken(ctx context.Context, params CloudsmithOIDCParamete
 	}
 
 	if tokenResp.Token == "" {
-		return nil, fmt.Errorf("Cloudsmith token response does not contain a token")
+		return nil, fmt.Errorf("cloudsmith token response does not contain a token")
 	}
 
 	// Cloudsmith tokens are valid for 2 hours according to their documentation
--- a/internal/oidc/actions_oidc.go
+++ b/internal/oidc/actions_oidc.go
@@ -573,9 +573,6 @@ func GetCloudsmithAccessToken(ctx context.Context, params CloudsmithOIDCParamete
 	if params.ApiHost == "" {
 		return nil, fmt.Errorf("API host is required")
 	}
-	if params.Audience == "" {
-		return nil, fmt.Errorf("audience is required")
-	}
 	if params.OrgName == "" {
 		return nil, fmt.Errorf("org name is required")
 	}
--- a/internal/oidc/actions_oidc.go
+++ b/internal/oidc/actions_oidc.go
@@ -178,6 +178,15 @@ type awsTokenResponse struct {
 	Expiration         float64 `json:"expiration"`
 }
 
+type cloudsmithTokenRequest struct {
+	OIDCToken   string `json:"oidc_token"`
+	ServiceSlug string `json:"service_slug"`
+}
+
+type cloudsmithTokenResponse struct {
+	Token string `json:"token"`
+}
+
 // OIDCAccessToken represents an access token with its expiry information
 type OIDCAccessToken struct {
 	Token     string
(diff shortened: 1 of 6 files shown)


internal/oidc/oidc_credential.go
```diff
@@ -15,6 +16,14 @@ import (
 	"github.com/dependabot/proxy/internal/logging"
 )
 
+const (
+	envActionsRepositoryOwner = "ACTIONS_REPOSITORY_OWNER"
+)
+
+func GetRepositoryOwner() string {
```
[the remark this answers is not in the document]
joniumGit: Renamed to correct variable

internal/oidc/actions_oidc.go
```diff
@@ -557,6 +566,96 @@ func GetAWSAccessTokenForDevOps(ctx context.Context, params AWSOIDCParameters) (
...
+func GetCloudsmithAccessToken(ctx context.Context, params CloudsmithOIDCParameters, githubToken string) (*OIDCAccessToken, error) {
+	if params.ServiceSlug == "" {
+		return nil, fmt.Errorf("service slug is required")
+	}
+	if params.ApiHost == "" {
+		return nil, fmt.Errorf("API host is required")
+	}
+	if params.Audience == "" {
+		return nil, fmt.Errorf("audience is required")
+	}
```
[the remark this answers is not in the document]
joniumGit: Removed the validation

internal/oidc/actions_oidc_test.go
```diff
@@ -832,3 +833,184 @@ func TestGetAWSAccessToken(t *testing.T) {
...
+			cloudsmithToken, err = GetCloudsmithAccessToken(ctx, tt.params, tt.githubToken)
+
+			if tt.expectError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+				require.NotNil(t, cloudsmithToken)
+				assert.Equal(t, tt.expectedToken, cloudsmithToken.Token)
+				assert.NotNil(t, cloudsmithToken.ExpiresIn)
+			}
```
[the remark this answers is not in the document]
joniumGit: Removed the assertion

internal/oidc/actions_oidc.go
```diff
@@ -557,6 +566,96 @@ func GetAWSAccessTokenForDevOps(ctx context.Context, params AWSOIDCParameters) (
...
+	requestBodyJson, err := json.Marshal(requestBody)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal cloudsmith token request: %w", err)
+	}
+
+	tokenURL := fmt.Sprintf("https://%s/openid/%s/", params.ApiHost, params.OrgName)
+	req, err := http.NewRequestWithContext(ctx, "POST", tokenURL, bytes.NewReader(requestBodyJson))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cloudsmith token request: %w", err)
+	}
```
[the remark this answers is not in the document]
joniumGit: This seems to be an established pattern in the codespace, so for consistency I will ignore this suggestion.

internal/oidc/oidc_credential.go
```diff
@@ -119,6 +143,25 @@ func CreateOIDCCredential(cred config.Credential) (*OIDCCredential, error) {
...
+		apiHost := cred.GetString("api-host")
+		if apiHost == "" {
+			apiHost = "api.cloudsmith.io"
+		}
+		audience := cred.GetString("audience")
+		if audience == "" {
+			if owner := GetRepositoryOwner(); owner != "" {
+				audience = fmt.Sprintf("https://github.com/%s", owner)
+			} else {
+				return nil, fmt.Errorf("missing audience for cloudsmith")
```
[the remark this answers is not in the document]
joniumGit: Made error message more clear.

internal/oidc/oidc_credential.go
```diff
@@ -15,6 +16,14 @@ import (
 	"github.com/dependabot/proxy/internal/logging"
 )
 
+const (
+	envActionsRepositoryOwner = "GITHUB_REPOSITORY_OWNER"
```
brettfo: The proxy won't have access to this environment variable so it can't be depended upon.  The only env vars that get forwarded to the proxy are defined [here](https://github.com/github/dependabot-action/blob/2c339a9695a8285e56f36eef47ae73bf772509d7/src/proxy.ts#L228).  Instead I'd make the `audience` parameter required.

internal/oidc/oidc_credential_test.go
```diff
@@ -268,16 +268,53 @@ func TestTryCreateOIDCCredential(t *testing.T) {
...
+			nil,
+		},
 	}
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			// these variables are necessary
 			os.Setenv(envActionsIDTokenRequestURL, "https://example.com/token")
 			os.Setenv(envActionsIDTokenRequestToken, "test-token")
+			os.Setenv(envActionsRepositoryOwner, "my-repo-owner")
```
brettfo: Same comment as above about this variable not being available when the proxy is started.

brettfo (changes requested):
Thank you for the PR!  Overall it looks great and exactly what I'd hope it would look like.  The only feedback (specific comments at their appropriate locations) is that the environment variable `GITHUB_REPOSITORY_OWNER` won't be available to the proxy at runtime so you'll have to make the `audience` parameter required.

Once this is all done there will be some internal PRs required to update the `dependabot.yml` schema to allow these additional values as well as some documentation updates, but I'll take care of that when the time comes.

internal/oidc/actions_oidc.go
```diff
@@ -557,6 +566,93 @@ func GetAWSAccessTokenForDevOps(ctx context.Context, params AWSOIDCParameters) (
...
+		return nil, fmt.Errorf("failed to execute cloudsmith token request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read cloudsmith token response body: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
brettfo: Could you expand this check to also allow `http.StatusCreated`?  The GitHub [actions implementation accepts](https://github.com/cloudsmith-io/cloudsmith-cli-action/blob/d9be6867d8be2d923ffb88ce3d082f9c2a89a62f/src/oidc-auth.js#L232) both 200 and 201 and `.StatusCreated` is what [corresponds to that](https://go.dev/src/net/http/status.go).

joniumGit: That's a good catch, I missed that detail. I will update it.

internal/oidc/actions_oidc.go
```diff
@@ -557,6 +566,93 @@ func GetAWSAccessTokenForDevOps(ctx context.Context, params AWSOIDCParameters) (
...
+	}
+
+	tokenURL := fmt.Sprintf("https://%s/openid/%s/", params.ApiHost, params.OrgName)
+	req, err := http.NewRequestWithContext(ctx, "POST", tokenURL, bytes.NewReader(requestBodyJson))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cloudsmith token request: %w", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
```
joniumGit, line 600: I add an extra header here compared to the CLI, but I don't think it should cause issues. The CLI implementation assumes the response to be JSON.

internal/oidc/actions_oidc.go
```diff
@@ -557,6 +566,93 @@ func GetAWSAccessTokenForDevOps(ctx context.Context, params AWSOIDCParameters) (
...
+	}
+
+	var tokenResp cloudsmithTokenResponse
+	if err := json.Unmarshal(body, &tokenResp); err != nil {
+		return nil, fmt.Errorf("failed to parse cloudsmith token response: %w", err)
+	}
+
+	if tokenResp.Token == "" {
+		return nil, fmt.Errorf("cloudsmith token response does not contain a token")
+	}
```
joniumGit, line 628: The cli seems to be more defensive in this validation, should I add more?

```typescript
!token || typeof token !== "string" || token.trim() === ""
```

joniumGit:
I went through the implementation again side by side with the CLI and docs after your comment. I only found a couple minor details I commented on.

internal/oidc/actions_oidc.go
brettfo: The current check against an empty string should suffice.  The deserialization specified in the struct ensures it's a string and if we're getting tokens with leading or trailing whitespace that seems like an issue with their endpoint.

brettfo (approved):

joniumGit:
Hello, I tested it this morning with python packages and was able to successfully use the OIDC auth to get a Dependabot PR to pop up for a private package.

The config I used was pretty much like this. The audience I set to match the example [in the OIDC Github docs for actions](https://docs.github.com/en/actions/concepts/security/openid-connect#understanding-the-oidc-token) so that I can use the same OIDC provider config in Cloudsmith for both.
```yaml
version: 2
registries:
  cloudsmith-python:
    replaces-base: true
    type: python-index
    # https://docs.cloudsmith.com/formats/python-repository#download--install-a-package
    url: https://dl.cloudsmith.io/public/OWNER/REPOSITORY/python/simple/
    oidc-audience: https://github.com/GITHUB-ORG
    oidc-namespace: OWNER
    oidc-service-slug: ${{ secrets.CLOUDSMITH_USERNAME }}
updates:
  - package-ecosystem: pip
    # One package needs this
    insecure-external-code-execution: allow
    registries: '*'
    directory: /
    schedule:
      interval: weekly
```

joniumGit:
Also tested NPM, Docker, and Maven to be working.

brettfo:
@joniumGit Thank you for checking!  I'll work on getting the documentation published with a note that the audience should look like you specified.  It might take a bit to get the docs up since there will likely be a blog post to go along with it, but in the meantime enjoy your secure package feeds.

Thanks again for all your work, we really appreciate it!

joniumGit:
Thank you!

brettfo:
@joniumGit (and anybody else following along): We're doing a refactor of the proxy and we're renaming the values for Cloudsmith to remove the `oidc-` prefix.  More details are in PR #99 but the short version is that the following properties should be renamed in your `dependabot.yml` file:

- `oidc-namespace` to `namespace`
- `oidc-service-slug` to `service-slug`
- `oidc-audience` to `audience`

We've made a sneaky change on the back end to auto-rename the `oidc-` values as they're read from the config file so hopefully you won't notice any downtime, but this auto-rename will be removed soon and you'll need to be using the new property names.

When you get a chance could you update your `dependabot.yml` property names to the above and let me know if it still works?  Once you've migrated to the new values we'll pull out the auto-renamer on the back end to invalidate the `oidc-` prefix and publish the documentation.

joniumGit:
Hi, thanks for the info. I will begin migrating all the repos that already use it.

PR #99, opened 2026-04-14
[Task] Refactor CloudSmith OIDC credential keys to remove oidc- prefix

kbukum1:

Refactor CloudSmith OIDC credential field names to remove the `oidc-` prefix for consistency with other OIDC providers (Azure, AWS, JFrog).

Changes:
- `oidc-namespace` → `namespace`
- `oidc-service-slug` → `service-slug`
- `oidc-audience` → `audience`

This also adds a PR template modeled after [dependabot-core](https://github.com/dependabot/dependabot-core/blob/main/.github/pull_request_template.md) to standardize PR descriptions.

The credential key names (`namespace`, `service-slug`, `audience`) are now consistent with how other providers define their OIDC parameters — without a provider-specific prefix. This is a breaking change for any configuration currently using the `oidc-` prefixed keys, so companion PRs in github/github and github-ui must land together.

### How will you know you have accomplished your goal?

All existing tests pass with the updated key names, confirming functional equivalence. The companion PRs in github/github and github-ui will update the configuration layer to emit the new key names.

### Checklist

### DIFF
--- a/internal/oidc/oidc_credential.go
+++ b/internal/oidc/oidc_credential.go
@@ -98,9 +98,9 @@ func CreateOIDCCredential(cred config.Credential) (*OIDCCredential, error) {
 	domainOwner := cred.GetString("domain-owner")
 
 	// cloudsmith values
-	orgName := cred.GetString("oidc-namespace")
-	serviceSlug := cred.GetString("oidc-service-slug")
-	cloudsmithAudience := cred.GetString("oidc-audience")
+	orgName := cred.GetString("namespace")
+	serviceSlug := cred.GetString("service-slug")
+	cloudsmithAudience := cred.GetString("audience")
 
 	switch {
 	case tenantID != "" && clientID != "":
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,27 @@
+### What are you trying to accomplish?
+
+<!-- Provide both a what and a _why_ for the change. -->
+
+<!-- What issues does this affect or fix? -->
+
+### Anything you want to highlight for special attention from reviewers?
+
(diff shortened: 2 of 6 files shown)


internal/oidc/oidc_credential.go
```diff
@@ -98,9 +98,9 @@ func CreateOIDCCredential(cred config.Credential) (*OIDCCredential, error) {
 	domainOwner := cred.GetString("domain-owner")
 
 	// cloudsmith values
-	orgName := cred.GetString("oidc-namespace")
-	serviceSlug := cred.GetString("oidc-service-slug")
-	cloudsmithAudience := cred.GetString("oidc-audience")
+	orgName := cred.GetString("namespace")
+	serviceSlug := cred.GetString("service-slug")
+	cloudsmithAudience := cred.GetString("audience")
```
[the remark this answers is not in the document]
kbukum1: We don't need fallback since it is not GA yet.

brettfo, line 103: I realize it's not GA yet (as per your reply to the copilot comment) but there is at least one person using this in their production environment and I don't want to break them.  Can we have a few days where both are allowed?  I don't think we need explicit test coverage for it but change this line and the others to:

``` go
orgName := cred.GetString("namespace")
if orgName == "" {
    orgName = cred.GetString("oidc-namespace")
}
```

brettfo (changes requested):

internal/oidc/oidc_credential.go
brettfo: Discussed elsewhere, the plan is to keep this change as-is and for a short time do a manual conversion of `oidc-namespace` to `namespace`, etc. on the back so by the time the proxy gets it, it's always the new value.

brettfo (approved):

PR #104, opened 2026-04-20
fix(cache): skip malformed cached responses to avoid invalid status line

thavaahariharangit:

Prevent Dependabot Cargo updates from failing with Weird server reply (Invalid status line) when fetching `crates.io` index entries through the proxy.

**Why:**
A malformed cached HTTP response could be replayed with invalid status metadata, which can produce an invalid HTTP status line for `Cargo/libcurl`. This caused update jobs to fail while downloading dependencies from `index.crates.io`.

What this change does:
Validates cached response `status/metadata` before replay, skips corrupt cache entries, and falls back to upstream fetch instead of returning malformed cached responses. Also adds a regression test to ensure invalid-status cache entries are ignored.

1. Automated verification
- `handlers_test.go:127`passes, proving corrupt cache entries with invalid status are ignored.
- go test `./internal/cache` passes (no regression in cache behavior).
2. Behavioral verification (original failure mode)
- Re-running the previously failing Dependabot Cargo update through this proxy no longer returns Weird server reply (Invalid status line) for `crates.io index` downloads.
3. Runtime safety verification
- When a malformed cache entry is encountered, proxy logs show it is skipped (not replayed), and requests fall back to upstream successfully.
- No functional regression
Valid cached responses continue to be served normally (existing cache hit path remains intact), while only malformed cached metadata is rejected.

### Checklist

### DIFF
--- a/internal/cache/handlers.go
+++ b/internal/cache/handlers.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"net/http/httputil"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -180,6 +181,11 @@ func (d *DB) OnRequest(r *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *
 	key := key(r)
 	ctxdata.SetValue(ctx, keyValue, key)
 	if entry, ok := d.cacheDB[key]; ok {
+		if http.StatusText(entry.Status) == "" {
+			logrus.Warnf("skipping cache entry with invalid status code (%d) for %s", entry.Status, r.URL.String())
+			return r, nil
+		}
+
 		f, err := os.Open(entry.FilePath)
 		if err != nil {
 			logrus.Errorln("failed to open cache file:", err)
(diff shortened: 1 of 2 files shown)


internal/cache/handlers.go
```diff
@@ -192,7 +198,18 @@ func (d *DB) OnRequest(r *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *
 		resp.TransferEncoding = r.TransferEncoding
 		resp.Header = entry.ResponseHeaders
 		resp.StatusCode = entry.Status
+		resp.Status = fmt.Sprintf("%d %s", entry.Status, http.StatusText(entry.Status))
+		resp.Proto = "HTTP/1.1"
+		resp.ProtoMajor = 1
+		resp.ProtoMinor = 1
 		resp.Body = f
+
+		if _, err := httputil.DumpResponse(resp, false); err != nil {
```
jakecoffman, line 207: This doesn't seem like a very strong validation of the response and it has a cost of reading the entire response into memory. I recommend not doing this part.

internal/cache/handlers.go
```diff
@@ -192,7 +198,18 @@ func (d *DB) OnRequest(r *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *
 		resp.TransferEncoding = r.TransferEncoding
 		resp.Header = entry.ResponseHeaders
 		resp.StatusCode = entry.Status
+		resp.Status = fmt.Sprintf("%d %s", entry.Status, http.StatusText(entry.Status))
+		resp.Proto = "HTTP/1.1"
+		resp.ProtoMajor = 1
+		resp.ProtoMinor = 1
```
jakecoffman, line 204: My guess is this is the actual fix because I tested with DumpResponse in a test and found it is setting HTTP/0.0 which I could see failing validation.

internal/cache/handlers.go
```diff
@@ -180,6 +181,11 @@ func (d *DB) OnRequest(r *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *
 	key := key(r)
 	ctxdata.SetValue(ctx, keyValue, key)
 	if entry, ok := d.cacheDB[key]; ok {
+		if http.StatusText(entry.Status) == "" {
```
jakecoffman, line 184: If the server uses custom HTTP methods that Go doesn't know about this will make all calls uncached. Not a big deal but I doubt this is the actual fix.

thavaahariharangit:
I am closing this update as the root cause has been resolved with reverting [this pr update](https://github.com/dependabot/proxy/pull/105)

PR #108, opened 2026-04-21
 Add GCP OIDC token exchange for Google Artifact Registry

kbukum1:

Add Google Artifact Registry as the **fifth OIDC provider** in the Dependabot proxy, alongside Azure DevOps, AWS CodeArtifact, JFrog, and Cloudsmith. The proxy mints a short-lived Google access token from the GitHub Actions OIDC JWT and injects it on outbound registry requests — no static credentials stored.

Token exchange flow:
- GitHub Actions OIDC JWT → Google STS token exchange
- Optional IAM Credentials `generateAccessToken` impersonation (when `service-account` is configured)
- Direct Workload Identity Federation when no `service-account` is present

Auth injection:
- `Authorization: Bearer <token>` for most registry types (Maven, npm, Python, etc.)
- `Basic oauth2accesstoken:<token>` for Docker hosts (`*-docker.pkg.dev`)

- Follows the Cloudsmith OIDC pattern: parameter struct, factory branch, token exchange function, ForDevOps wrapper, and header injection in OIDCRegistry.TryAuth.
- The Docker Basic auth pattern (`oauth2accesstoken:<token>`) is new — all 4 existing providers use Bearer or X-Api-Key. GCP Artifact Registry's Docker API requires this specific format.
- STS request uses **camelCase** JSON (Google's gRPC-transcoded convention), not RFC 8693 snake_case. The response uses snake_case.
- No new Go dependencies — uses raw `net/http` matching existing providers.

### How will you know you have accomplished your goal?

- `go test ./... -count=1` — all packages pass
- `go vet ./...` — clean
- `go build ./...` — succeeds

### Checklist

### DIFF
--- a/internal/handlers/oidc_handling_test.go
+++ b/internal/handlers/oidc_handling_test.go
@@ -133,6 +133,27 @@ func TestOIDCURLsAreAuthenticated(t *testing.T) {
 				"https://cloudsmith.example.com/some-package",
 			},
 		},
+		{
+			name:     "Cargo",
+			provider: "gcp",
+			handlerFactory: func(creds config.Credentials) oidcHandler {
+				return NewCargoRegistryHandler(creds)
+			},
+			credentials: config.Credentials{
+				config.Credential{
+					"type":                       "cargo_registry",
+					"url":                        "https://us-central1-cargo.pkg.dev/my-project/my-repo",
+					"workload-identity-provider": "projects/123/locations/global/workloadIdentityPools/pool/providers/prov",
+				},
+			},
+			urlMocks: []mockHttpRequest{},
+			expectedLogLines: []string{
+				"registered gcp OIDC credentials for cargo registry: https://us-central1-cargo.pkg.dev/my-project/my-repo",
+			},
+			urlsToAuthenticate: []string{
+				"https://us-central1-cargo.pkg.dev/my-project/my-repo/some-package",
+			},
+		},
 		//
 		// Composer
 		//
@@ -226,6 +247,27 @@ func TestOIDCURLsAreAuthenticated(t *testing.T) {
 			},
 			urlsToAuthenticate: []string{
 				"https://cloudsmith.example.com/some-package",
+			},
+		},
+		{
+			name:     "Composer",
+			provider: "gcp",
+			handlerFactory: func(creds config.Credentials) oidcHandler {
+				return NewComposerHandler(creds)
+			},
+			credentials: config.Credentials{
+				config.Credential{
+					"type":                       "composer_repository",
+					"registry":                   "https://us-central1-composer.pkg.dev/my-project/my-repo",
+					"workload-identity-provider": "projects/123/locations/global/workloadIdentityPools/pool/providers/prov",
+				},
+			},
+			urlMocks: []mockHttpRequest{},
+			expectedLogLines: []string{
+				"registered gcp OIDC credentials for composer repository: https://us-central1-composer.pkg.dev/my-project/my-repo",
+			},
+			urlsToAuthenticate: []string{
+				"https://us-central1-composer.pkg.dev/my-project/my-repo/some-package",
 			},
 		},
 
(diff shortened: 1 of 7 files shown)


kbukum1:
## Follow-up: Commonize OIDC handler test cases

While working on this PR, I noticed that `internal/handlers/oidc_handling_test.go` has a highly repetitive structure — each ecosystem (Cargo, Composer, NPM, Maven, Python, NuGet, Hex, Pub, RubyGems, Go, Docker, Helm, Terraform) repeats nearly identical test blocks for each OIDC provider (AWS, Azure, JFrog, Cloudsmith, GCP), totaling ~65 hand-written cases.

This could be collapsed with a table-driven approach: define each ecosystem's metadata once (handler factory, credential type, URL key, log label, URLs), then generate provider variants in a loop — reducing to ~13 ecosystem entries + a shared template (~75% less test code, same coverage).

Suggesting this as a follow-up refactor to keep this PR focused.

kbukum1:
## Follow-up: Extract shared `ForDevOps` wrapper

All 5 `Get*AccessTokenForDevOps` functions (`actions_oidc.go:272,373,549,637,817`) repeat the same flow:
1. `IsOIDCConfigured()` check
2. `GetToken(ctx, audience)` (Azure uses `GetTokenForAzureADExchange`)
3. `GetXxxAccessToken(ctx, params, githubToken)`
4. Wrap error with provider name

A shared helper with a `tokenExchangeFn` callback could eliminate ~15 lines per provider. Worth evaluating as a follow-up refactor.

kbukum1:
## Follow-up: Extract JSON POST helpers for OIDC token exchange

All OIDC providers (Azure, JFrog, AWS, Cloudsmith, GCP) repeat the same ~15-line HTTP boilerplate: marshal → create request → set standard headers → execute → read body. GCP does it twice (STS + IAM).

### Proposed approach: build/execute split

Rather than a single helper with growing parameter lists, split into two:

```go
// buildJSONRequest creates a JSON POST request with standard proxy headers.
// Caller can modify the request (add headers, etc.) before executing it.
func buildJSONRequest(ctx context.Context, url string, body any) (*http.Request, error) {
    bodyJSON, err := json.Marshal(body)
    if err != nil {
        return nil, fmt.Errorf("failed to marshal request: %w", err)
    }
    req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewReader(bodyJSON))
    if err != nil {
        return nil, fmt.Errorf("failed to create request: %w", err)
    }
    req.Header.Set("Content-Type", "application/json")
    req.Header.Set("Accept", "application/json")
    req.Header.Set("User-Agent", "dependabot-proxy/1.0")
    return req, nil
}

// executeJSONRequest executes a request and returns the status code and raw body.
func executeJSONRequest(req *http.Request) (int, []byte, error) {
    client := &http.Client{Timeout: 10 * time.Second}
    // ... execute, read body, return statusCode, body, err
}
```

### Usage example (GCP IAM with custom auth header):

```go
req, err := buildJSONRequest(ctx, iamURL, iamReqBody)
if err != nil { ... }
req.Header.Set("Authorization", "Bearer "+stsToken)  // custom header

status, body, err := executeJSONRequest(req)
if err != nil { ... }
```

### Why this design:
- **`buildJSONRequest`** standardizes the proxy OIDC conventions (JSON content type, User-Agent, timeout)
- **Caller has full control** between build and execute — add any headers, query params, or modifications without changing the helper
- **Provider-specific logic stays inline** — status checks, error context, response parsing remain visible at the call site
- **No parameter proliferation** — future needs don't require helper signature changes

kbukum1:
## Follow-up: Consider provider interface to replace type switches

Adding a new OIDC provider currently requires synchronized edits in 3 places:
1. `oidc_credential.go:114-168` — parse config fields into params struct
2. `oidc_credential.go:202-214` — type switch to dispatch token exchange
3. `oidc_registry.go:143-158` — type switch to apply auth to requests

An `OIDCProvider` interface with `ExchangeToken()` and `ApplyAuth()` methods could eliminate switches #2 and #3, reducing new-provider additions to a single edit point (config parsing). Worth considering if more providers are expected.

kbukum1:
## Follow-up: Simplify `oidc_credential_test.go` parameter assertions

The type switch at `oidc_credential_test.go:388-436` does field-by-field comparison of simple value structs across all 5 providers (~50 lines). Since these parameter types have no unexported fields or special comparison logic, the entire block could be replaced with:

```go
assert.Equal(t, tc.expectedParameters, actual.parameters)
```

`assert.Equal` uses `reflect.DeepEqual`, which compares all fields and gives clear diffs on failure. The type check is already covered by line 386. This would reduce 50 lines to 1 line with the same coverage.

AbhishekBhaskar (approved):

PR #115, opened 2026-04-29
Add OpenTofu registry handler

diofeher:
## Summary

- dependabot-core's OpenTofu ecosystem accepts credentials with `type: opentofu_registry`, but the proxy had no matching handler. Requests against private OpenTofu registries reached the upstream anonymously and failed authentication.
- Adds `OpenTofuRegistryHandler` in `internal/handlers/opentofu_registry.go` modeled on the existing Terraform handler. The wire protocol is identical (Terraform Module Registry HTTP API), so the logic is the same — only the credential-type filter, log message, and OIDC label are renamed.
- Registers the new handler in `proxy.go` after the Terraform handler.


v-thavaahariharangit:
**Required follow‑up work** 

- **Dependabot API schema update** — The PR introduces a new registry type `type: opentofu_registry`. We need to extend the Dependabot API schema to recognize and validate this new registry type. 

- **Dependabot-core module analysis** — Investigate whether private registry support for OpenTofu already exists in dependabot-core, and identify any missing pieces.

- **Integration testing** — Add a full integration test to verify private registry authentication and end‑to‑end functionality.

thavaahariharangit (approved):

PR #124, opened 2026-05-14
fix(dialer): enforce IP block list before socket creation

yeikel:

The blocked-IP check in `safeControl` runs after the OS creates the socket. On systems where a network family is unsupported (e.g. IPv6 disabled), socket creation fails first with "address family not supported by protocol" `safeControl` is never reached, so `ErrForbiddenRequest` is never returned and `handleForbidden` can't produce a 403.

Fix by checking resolved IPs against the block list inside `DialContext`, before `net.Dialer.DialContext` is called. `safeControl` is kept as a defence-in-depth fallback for any path that bypasses `DialContext`.

Fixes `TestIPRestrictions/http://[::1]/` on CI runners without IPv6.

This only happens on self hosted runners that do not support IPv6 so it is difficult to provide a reproducer or test here

Hosts without IPV6 support continue to work

### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.26.0
 
 require (
 	github.com/aws/aws-sdk-go v1.55.8
-	github.com/aws/aws-sdk-go-v2 v1.41.7
+	github.com/aws/aws-sdk-go-v2 v1.41.12
 	github.com/elazarl/goproxy v1.8.4
 	github.com/evalphobia/logrus_sentry v0.8.2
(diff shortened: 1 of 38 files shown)


yeikel:
I just noticed that there are some test failures but these failures are just related to https://github.com/dependabot/smoke-tests/issues/277

yeikel:
I re-generated the tests here https://github.com/dependabot/smoke-tests/pull/522

JamieMagee (approved):

PR #128, opened 2026-05-21
Unify safe dialer use

joniumGit:
_This PR is co-authored by Copilot_

I am trying to unify the safeDialer use in the codebase. I noticed that there were plenty of user input going into Http Clients in some parameters and them not using the safe dialer that the proxy explicitly creates in the beginning. This will prevent any config parameters from being used to reach unintended destinations. The way I did it is a bit ugly and ends up with the transport being added to all of the registry handler constructors.

Please check if this change is needed. I am not sure if the initial state of not using safe dialer in some of the registry handlers was intended.

No user input should reach a raw HTTP Client.

### Checklist

### DIFF
--- a/internal/oidc/actions_oidc.go
+++ b/internal/oidc/actions_oidc.go
@@ -602,6 +602,9 @@ func GetCloudsmithAccessToken(ctx context.Context, params CloudsmithOIDCParamete
 	if params.OrgName == "" {
 		return nil, fmt.Errorf("org name is required")
 	}
+	if params.Audience == "" {
+		return nil, fmt.Errorf("audience is required")
+	}
 	if githubToken == "" {
 		return nil, fmt.Errorf("GitHub token is required")
 	}
--- a/internal/oidc/actions_oidc.go
+++ b/internal/oidc/actions_oidc.go
@@ -227,7 +227,7 @@ type OIDCAccessToken struct {
 // githubToken: The GitHub Actions OIDC token obtained via GetTokenForAzureADExchange
 //
 // Returns an Azure AD access token scoped for Azure DevOps (499b84ac-1321-427f-aa17-267ca6975798/.default)
-func GetAzureAccessToken(ctx context.Context, params AzureOIDCParameters, githubToken string) (*OIDCAccessToken, error) {
+func GetAzureAccessToken(ctx context.Context, params AzureOIDCParameters, githubToken string, client *http.Client) (*OIDCAccessToken, error) {
 	if params.TenantID == "" {
 		return nil, fmt.Errorf("tenant ID is required")
 	}
@@ -250,10 +250,6 @@ func GetAzureAccessToken(ctx context.Context, params AzureOIDCParameters, github
 	formData.Set("client_assertion_type", "urn:ietf:params:oauth:client-assertion-type:jwt-bearer")
 	formData.Set("client_assertion", githubToken)
 	formData.Set("grant_type", "client_credentials")
-
-	client := &http.Client{
-		Timeout: 10 * time.Second,
-	}
 
 	req, err := http.NewRequestWithContext(ctx, "POST", tokenURL, strings.NewReader(formData.Encode()))
 	if err != nil {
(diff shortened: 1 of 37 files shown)


internal/oidc/actions_oidc.go
```diff
@@ -227,7 +227,7 @@ type OIDCAccessToken struct {
 // githubToken: The GitHub Actions OIDC token obtained via GetTokenForAzureADExchange
 //
 // Returns an Azure AD access token scoped for Azure DevOps (499b84ac-1321-427f-aa17-267ca6975798/.default)
-func GetAzureAccessToken(ctx context.Context, params AzureOIDCParameters, githubToken string) (*OIDCAccessToken, error) {
+func GetAzureAccessToken(ctx context.Context, params AzureOIDCParameters, githubToken string, client *http.Client) (*OIDCAccessToken, error) {
```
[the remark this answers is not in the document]
joniumGit: These are now unified in the call before this where the actual client is created. This alert is only half-correct in my opinion, there could be a nil check for the value.

jeffwidman:
@joniumGit 👋 Sorry this took a while for us to get to... 

I started to review when I noticed merge conflicts and also a bunch of agent review comments... 

If you can rebase and then take a look at those--don't have to agree, but at least take a look and if you disagree note why, after that I'd be happy to review.

jeffwidman (commented):
> Agent note: This comment was produced by an agent after an adversarial review, guided by Jeff on the specific areas to examine. It is not Jeff speaking directly, so maintainers may hold these conclusions less strongly.

- The underlying problem is worth solving: OIDC exchanges can carry assertions or generated credentials, and bypassing the proxy's restricted transport weakens an intentional SSRF boundary. This is meaningful defense in depth, though currently narrow: production blocks cloud metadata destinations, not private networks generally, and exploitation appears to require influence over job credentials or OIDC configuration.
- The implementation is directionally sound, but incomplete. Provider token exchanges use the injected transport, while the initial GitHub Actions `GetToken` request still uses a raw client; the ECR SDK client also remains outside this boundary unless that is explicitly considered trusted.
- Construct one `*http.Client` at the composition root with the safe transport and the existing bounded timeout, then inject that client through the OIDC registry/credential path. Passing the client, rather than a nullable `RoundTripper`, keeps timeout and transport policy together and avoids silent fallback to `http.DefaultTransport`. Production constructors should reject or make `nil` impossible; tests should pass an explicit test client.
- Use that same client for every leg of the flow: the GitHub Actions `GetToken` request and each Azure/JFrog/AWS/Cloudsmith/GCP exchange. Redirects must remain on the safe transport as well. For ECR, supply the client through the AWS SDK HTTP-client option, or explicitly document why that endpoint is outside this boundary.
- Add focused regression coverage for the invariant rather than only adapting signatures with `nil` or `http.DefaultClient`: a recording `RoundTripper` should prove both OIDC legs use the injected client, and an integration-style test should prove direct and redirected requests to a blocked destination fail without making a connection. Retain an assertion for the 10-second timeout.
- The Cloudsmith audience validation looks reasonable, but is unrelated and would be clearer as a separate change.

Overall: preserve the idea, but make the restricted client a mandatory dependency, cover all outbound credential-bearing requests, and test the security boundary directly before merging.

internal/oidc/actions_oidc.go
```diff
@@ -118,8 +117,21 @@ func GetToken(ctx context.Context, audience string) (_ string, err error) {
...
+	if client == nil {
+		return fmt.Errorf("OIDC HTTP client is required")
+	}
+	if client.Transport == nil {
+		return fmt.Errorf("OIDC HTTP client transport is required")
+	}
+	if client.Timeout <= 0 {
+		return fmt.Errorf("OIDC HTTP client timeout must be positive")
+	}
+	return nil
```
joniumGit, line 134: I am thinking this could be just a nil check for the client

joniumGit:
I agree with the review feedback. 

Originally, I left the Github Actions and AWS SDK clients untouched as I was not sure if they need to reach some SafeDialer blocked destination. I also used the transport as a parameter instead of a client to minimize changes as that pattern was used already.

Anyways, I made Copilot rebase the changes, change the parameter to a HttpClient, and add tests that your review flagged. I can separate the audience validation to another PR, but I think it is a small change overall and could go as part of this as well, it is in a separate commit.

joniumGit:
Oh right, the Copilot review comments. They are all about a `nil` check for the HttpClient in the OIDC code. I personally don't see it necessary to nil check that as I don't think there is any valid reason to pass null to those functions, but I am open to other opinions.

internal/oidc/actions_oidc.go
jeffwidman: It doesn't hurt to be defensive I guess

jeffwidman (approved):

jeffwidman:
Thanks!

This is a nice security cleanup / hardening both of existing code but also going forward.

PR #134, opened 2026-06-03
Skip unroutable IP families before dialing

JamieMagee:

On IPv4-only hosts the proxy still tried IPv6 addresses returned by DNS, hung until the socket timed out, and that cascaded into the updater's 100s `HttpClient.Timeout`. Now we drop addresses whose family the host can't route before dialing, so single-stack hosts fail fast. IPv6-only hosts had the same problem with IPv4 addresses, so both families get the same treatment.

The probe is a connected UDP dial to a documentation address (`192.0.2.1` / `2001:db8::1`). It checks for a route, not real reachability, and caches the result for the process lifetime. A blackhole route could still report "available" (same caveat for both families).

`go test ./internal/dialer/` passes, including new cases: IPv4-only keeps only IPv6, IPv6-only keeps only IPv4, and neither available returns `ErrNoUsableAddress` without dialing.

### Checklist

### DIFF
--- a/internal/dialer/dialer.go
+++ b/internal/dialer/dialer.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"sync"
 	"syscall"
+	"time"
 
 	"github.com/rs/dnscache"
 )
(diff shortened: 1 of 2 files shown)


h5dh2gy26z-sys (approved):

internal/dialer/dialer.go
```diff
@@ -46,15 +65,129 @@ func (d *Dialer) DialContext(ctx context.Context, network, address string) (conn
...
+	ipv6AvailabilityOnce sync.Once
+	ipv6Availability     bool
+)
+
+// defaultIPv4Available reports whether the host can route IPv4 traffic. It
+// probes once and caches the result, since IPv4 availability won't change
+// while the process is running.
+func defaultIPv4Available() bool {
+	ipv4AvailabilityOnce.Do(func() {
+		ipv4Availability = checkConnectivity("udp4", "192.0.2.1:80")
```
brettfo, line 164: Exposing my own ignorance here: if the proxy isn't connected on the `192.168.*.*` local IPs but rather on `10.*.*.*` or any of the others, will this report an incorrect result, or is just the formatting of this as IPv4 (vs v6 below) sufficient?

JamieMagee: The `192.0.2.0/24` range is different from `192.168.0.0/16`. `192.0.2.0/24` is a test network that should always resolved. See https://en.wikipedia.org/wiki/List_of_reserved_IP_addresses

brettfo: Ah, I didn't even catch that, I just saw `192` and skipped to the end!

brettfo (approved):

yeikel:
I am guessing that this superseded and made https://github.com/dependabot/proxy/pull/124 redundant, right?

JamieMagee:
@yeikel ah, apologies. Yes, it looks like it. Is your issue resolved?

PR #139, opened 2026-06-09
Authenticate Python downloads discovered from Simple API responses

kbukum1:

Allow the proxy to authenticate Python package file downloads when an authenticated Simple API response points to a different stable download path than the configured index URL.

This is needed for Azure Artifacts Python indexes. Dependabot authenticates requests to the configured Simple API path, such as `/pypi/simple/...`, but Azure can return package links under rewritten `/pypi/download/...` paths that include internal project/feed identifiers. Those download URLs no longer match the original credential path, so hosted Dependabot can reach the package listing but fail to download the package file.

This affects dependabot/dependabot-core#15207.

The implementation intentionally keeps authentication scoped. It does not broaden Python index credentials to the whole host. Instead, the proxy only learns Azure Artifacts download prefixes from authenticated Python Simple API responses, then reuses the same basic or OIDC credential for later requests under those discovered prefixes.

Discovery is limited to successful Simple API HTML/JSON responses, bounded in memory, restricted to same-origin and same-organization download links, and capped so a long-running proxy cannot learn unbounded prefixes.

The proxy test coverage exercises Python Simple API download-prefix discovery for both static and OIDC-backed credentials, including scoped matching, oversized responses, custom ports, IPv6 hosts, and negative cases where discovery must not broaden auth.

Validation run:

- `go test ./internal/handlers ./internal/oidc`

### Checklist

### DIFF
--- a/internal/handlers/python_index_download_prefixes.go
+++ b/internal/handlers/python_index_download_prefixes.go
@@ -186,8 +186,9 @@ func normalizedPythonIndexDownloadURL(u *url.URL) *url.URL {
 
 	normalized := *u
 	normalized.Scheme = strings.ToLower(normalized.Scheme)
+	port := normalizedPort(&normalized)
 	normalized.Host = strings.ToLower(normalized.Hostname())
-	if port := normalizedPort(&normalized); !isDefaultPort(normalized.Scheme, port) {
+	if port != "" && !isDefaultPort(normalized.Scheme, port) {
 		normalized.Host += ":" + port
 	}
 	normalized.RawQuery = ""
--- a/internal/handlers/python_index_download_prefixes.go
+++ b/internal/handlers/python_index_download_prefixes.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"html"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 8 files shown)


internal/handlers/python_index_download_prefixes.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, false
+	}
+
+	parsedLink, err := url.Parse(link)
+	if err != nil {
+		return nil, false
+	}
+
+	resolved := baseURL.ResolveReference(parsedLink)
+	if resolved.Scheme != "https" || !sameOrigin(baseURL, resolved) || firstPathSegment(baseURL.Path) != firstPathSegment(resolved.Path) {
```
JamieMagee, line 159: There are three conditions here (https, same origin, same first path segment), but only the last one has a test behind it. Could you add some more test coverage please?

JamieMagee (dismissed):
I'd like a little bit more test coverage, but it's not a blocker, and everything else looks fine.

internal/handlers/python_index_download_prefixes.go
kbukum1: fixed

JamieMagee (approved):

PR #140, opened 2026-06-10
Add host and username/password support to cargo registry handler

brettfo:
The cargo handler previously only accepted url and token credential values. This adds support for:

- **host** as an alternative to url for matching requests (using helpers.CheckHost)
- **username/password** as an alternative to token for authentication (sets Basic auth)

When both token and username/password are provided, token takes precedence. Credentials with neither url nor host are rejected with a warning.

Modeled after the NuGet feed handler's approach to dynamically handling these credential variants.

## Tests added
- Host-only matching with token auth
- URL matching with username/password (basic auth)
- Host matching with username/password (basic auth)
- Token precedence over password
- Rejection of credentials with no url or host

### DIFF
--- a/internal/handlers/cargo_registry.go
+++ b/internal/handlers/cargo_registry.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/elazarl/goproxy"
 	"github.com/sirupsen/logrus"
@@ -39,8 +40,11 @@ type CargoRegistryHandler struct {
 }
 
 type cargoRepositoryCredentials struct {
-	url           string
-	authorization string
+	url      string
+	host     string
+	token    string
+	username string
+	password string
 }
 
 func NewCargoRegistryHandler(credentials config.Credentials) *CargoRegistryHandler {
@@ -55,6 +59,7 @@ func NewCargoRegistryHandler(credentials config.Credentials) *CargoRegistryHandl
 		}
 
 		url := credential.GetString("url")
+		host := strings.ToLower(credential.GetString("host"))
 
 		// Cargo credentials must remain URL-scoped; do not allow OIDC
 		// registration to fall back to host-only matching when url is empty.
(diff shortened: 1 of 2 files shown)


internal/handlers/cargo_registry.go
```diff
@@ -55,6 +59,7 @@ func NewCargoRegistryHandler(credentials config.Credentials) *CargoRegistryHandl
 		}
 
 		url := credential.GetString("url")
+		host := strings.ToLower(credential.GetString("host"))
```
kbukum1, line 62: Can we do scope `host` only when `url` is empty - aligns with Terraform.

internal/handlers/cargo_registry_test.go
```diff
@@ -96,3 +96,121 @@ func TestCargoRegistryHandler(t *testing.T) {
 	req = handleRequestAndClose(handler, req, nil)
 	assertUnauthenticated(t, req, "non-GET request")
 }
+
```
kbukum1, line 99: Can you validate and add regression test if we do the scoping case: credential with url (path-scoped) and host set → request to a different path on the same host should be unauthenticated?

internal/handlers/cargo_registry.go
```diff
@@ -67,18 +72,33 @@ func NewCargoRegistryHandler(credentials config.Credentials) *CargoRegistryHandl
...
+
+		if url != "" {
+			if _, err := helpers.ParseURLLax(cargoCred.url); err != nil {
+				logrus.Warnf("ignoring invalid registry url (%s): %v", cargoCred.url, err)
+				continue
+			}
+		} else if host == "" {
+			logrus.Warn("ignoring cargo_registry credential with no url or host")
 			continue
 		}
```
kbukum1, line 98:
```suggestion
		cargoCred := cargoRepositoryCredentials{
			url:      url,
			token:    token,
			username: username,
			password: password,
		}

		if url != "" {
			if _, err := helpers.ParseURLLax(cargoCred.url); err != nil {
				logrus.Warnf("ignoring invalid registry url (%s): %v", cargoCred.url, err)
				continue
			}
		} else if host != "" {
			// Only set host when url is empty so URL/path scoping always
			// takes precedence and never falls back to host-only matching.
			cargoCred.host = host
		} else {
			logrus.Warn("ignoring cargo_registry credential with no url or host")
			continue
		}
```

internal/handlers/cargo_registry.go
```diff
@@ -67,18 +72,33 @@ func NewCargoRegistryHandler(credentials config.Credentials) *CargoRegistryHandl
...
-		if cargoCred.authorization == "" {
-			logrus.Warnf("missing token for registry url (%s)", cargoCred.url)
+
+		if token == "" && password == "" {
+			logrus.Warnf("missing token or password for registry (url: %s, host: %s)", cargoCred.url, cargoCred.host)
```
kbukum1:
```suggestion
			logrus.Warnf("missing token or username/password for registry (url: %s, host: %s)", cargoCred.url, cargoCred.host)
```

internal/handlers/cargo_registry_test.go
```diff
@@ -96,3 +96,146 @@ func TestCargoRegistryHandler(t *testing.T) {
...
+	// matching url path should authenticate
+	req := httptest.NewRequest("GET", "https://cargo.example.com/registry/crate", nil)
+	req = handleRequestAndClose(handler, req, nil)
+	assertHasTokenAuth(t, req, "", token, "url-matched request")
+
+	// different path on same host should NOT authenticate (url scoping takes precedence)
+	req = httptest.NewRequest("GET", "https://cargo.example.com/other/path", nil)
+	req = handleRequestAndClose(handler, req, nil)
+	assertUnauthenticated(t, req, "different path on same host should not be authenticated when url is set")
+}
```
kbukum1, line 241:
```suggestion
}

func TestCargoRegistryHandlerUrlScopingNotBypassedByHost(t *testing.T) {
	token := "Bearer abc123" //nolint:gosec // test credential

	credentials := config.Credentials{
		config.Credential{
			"type":  "cargo_registry",
			"url":   "https://cargo.example.com/myorg",
			"host":  "cargo.example.com",
			"token": token,
		},
	}

	handler := NewCargoRegistryHandler(credentials)

	// in-scope path should authenticate
	req := httptest.NewRequest("GET", "https://cargo.example.com/myorg/crate", nil)
	req = handleRequestAndClose(handler, req, nil)
	assertHasTokenAuth(t, req, "", token, "in-scope path request")

	// different path on the same host must NOT authenticate
	req = httptest.NewRequest("GET", "https://cargo.example.com/otherorg/crate", nil)
	req = handleRequestAndClose(handler, req, nil)
	assertUnauthenticated(t, req, "host must not bypass url path scoping")
}
```

kbukum1 (approved):

PR #146, opened 2026-06-17
fix: add path-prefix matching for static npm registry credentials

### LINKED ISSUE #145: Static token credential collision when multiple npm registries share a host
## Problem

When two npm registries share the same hostname but use different URL paths, the proxy sends the wrong static token credential. The `HandleRequest` loop in `npm_registry.go` matches on host+port only and returns on the first match — it never compares the request URL path against the credential's registry path.

This is the static-token equivalent of #87 / #71, which were fixed for OIDC credentials via `OIDCRegistry` (longest path-prefix matching). The static credential fallback path was not updated.

## Reproduction

`dependabot.yml`:
```yaml
registries:
  registry-a:
    type: npm-registry
    url: https://artifactory.example.com/artifactory/api/npm/team-a-npm
    token: ${{secrets.TEAM_A_TOKEN}}
  registry-b:
    type: npm-registry
    url: https://artifactory.example.com/artifactory/api/npm/team-b-npm
    token: ${{secrets.TEAM_B_TOKEN}}
```

The proxy receives credentials in order: `[registry-a, registry-b]`. A request to `https://artifactory.example.com/artifactory/api/npm/team-b-npm/@scope/pkg` matches `registry-a` first (same host, same port) and sends `TEAM_A_TOKEN` → 403 from Artifactory because that token is scoped to `team-a-npm` only.

## Proxy logs

```
proxy | [056] GET https://artifactory.example.com:443/artifactory/api/npm/team-b-npm/@scope%2Fpkg
proxy | [056] * authenticating npm registry request (host: artifactory.example.com, token auth)
proxy | [056] 403 https://artifactory.example.com:443/artifactory/api/npm/team-b-npm/@scope%2Fpkg
```

## Root cause

`npm_registry.go` lines 80–105 — the static credential loop:

```go
for _, cred := range h.credentials {
    regURL, err := helpers.ParseURLLax(cred.registry)
    // ...
    if !npmRegistryHostMatches(host, reqHost) {
        continue
    }
    // No path check — first host+port match wins
    // ...
    return req, nil
}
```

## Suggested fix

Add path-prefix matching before applying the credential, consistent with `OIDCRegistry.TryAuth`:

```go
regPath := strings.TrimSuffix(regURL.Path, "/")
if regPath != "" && !strings.HasPrefix(req.URL.Path, regPath+"/") && req.URL.Path != regPath {
    continue
}
```

This ensures `/team-a-npm` credentials only apply to `/team-a-npm/...` requests.

## Impact

Any user with multiple npm registries on the same host using static tokens (non-OIDC). Common with JFrog Artifactory, AWS CodeArtifact, and Azure DevOps where a single hostname serves multiple virtual repositories.

## Related

- #87 — OIDC credential collision (fixed)
- #71 — Wrong OIDC token for npm same host (fixed)
- #78 — OIDCRegistry implementation
- dependabot/dependabot-core#15248 — updater-side path matching fix (does not help because proxy sends wrong token before updater logic runs)

ryemelia:
Fixes #145

## What

Add path-segment-aware matching to the static credential loop in `NPMRegistryHandler.HandleRequest`, so that credentials configured for one path-scoped registry are not applied to sibling paths on the same host.

## Why

When multiple npm registries share the same hostname but use different URL paths (common with JFrog Artifactory, AWS CodeArtifact, and Azure DevOps), the proxy sends the first matching credential based on host+port only. If the tokens are scoped to their respective repos, the wrong token results in a 403.

The OIDC code path was already fixed via `OIDCRegistry` (#78, #87, #91), but the static token fallback was not updated.

## How

After the existing host+port check, compare the request URL path against the credential's registry path using segment-aware prefix matching:

```go
regPath := strings.TrimSuffix(regURL.Path, "/")
if regPath != "" && !strings.HasPrefix(req.URL.Path, regPath+"/") && req.URL.Path != regPath {
    continue
}
```

- Empty path (host-only registries like `registry.npmjs.org`) continues to match any path on that host — no behavior change.
- Non-empty paths require the request URL to be within the registry's path scope.
- Segment-aware: `/team-a-npm` does not match `/team-a-npm-evil`.

## Tests

- Updated existing test: sibling path on same host is now correctly rejected.
- Added `TestNPMRegistryHandler_SameHostDifferentPaths`: verifies two registries on the same host with different paths receive their respective tokens, and unrelated paths are unauthenticated.

All existing tests pass (`go test ./...`).

## Checklist

### DIFF
--- a/internal/handlers/npm_registry.go
+++ b/internal/handlers/npm_registry.go
@@ -101,6 +101,14 @@ func (h *NPMRegistryHandler) HandleRequest(req *http.Request, ctx *goproxy.Proxy
 			continue
 		}
 
+		// Path-segment-aware matching prevents credentials configured for one
+		// path-scoped registry from being applied to sibling paths on the same
+		// host (e.g., /team-a-npm should not match /team-b-npm).
+		regPath := strings.TrimSuffix(regURL.Path, "/")
+		if regPath != "" && !strings.HasPrefix(req.URL.Path, regPath+"/") && req.URL.Path != regPath {
+			continue
+		}
+
 		if cred.token == "" && cred.password != "" {
 			cred.token = cred.username + ":" + cred.password
 		}
--- a/internal/handlers/npm_registry_test.go
+++ b/internal/handlers/npm_registry_test.go
@@ -59,9 +59,10 @@ func TestNPMRegistryHandler(t *testing.T) {
 	req = handleRequestAndClose(handler, req, nil)
 	assertHasTokenAuth(t, req, "Bearer", privateRegToken, "valid registry request with port and path")
 
+	// Sibling path on the same host should NOT receive credentials from /reg-path
 	req = httptest.NewRequest("GET", "https://example.com/other-path/private-package", nil)
 	req = handleRequestAndClose(handler, req, nil)
-	assertHasTokenAuth(t, req, "Bearer", privateRegToken, "different path")
+	assertUnauthenticated(t, req, "sibling path should not match")
 
 	req = httptest.NewRequest("GET", "https://nexus.some-company.com/private-package", nil)
 	req = handleRequestAndClose(handler, req, nil)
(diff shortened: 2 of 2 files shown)


jeffwidman (approved):
Makes sense to me, thank you

PR #176, opened 2026-07-30
Add JIT access fallback for GitHub API requests

thavaahariharangit:

Add JIT access fallback support to GitHubAPIHandler.

When static credentials cannot access a repository and GitHub returns `401`, `403`, or `404`, the handler now requests a repository-scoped JIT credential and retries the API request. This allows Dependabot to retrieve metadata, such as release information, from internal repositories permitted by the organization’s repository access policy.

Successful JIT credentials are cached for subsequent API requests to the same repository.

The implementation follows the existing JIT authentication pattern used by GitServerHandler:
- Static credentials are attempted first.
- JIT access is requested only after authentication fails.
- JIT credentials are scoped and cached per repository.
- JIT fallback only runs for requests that the proxy authenticated with static credentials.
- The existing API client is shared with `GitHubAPIHandler`.
Reviewers should pay particular attention to treating `404` as a potential authentication failure, since GitHub returns it when an authenticated token cannot access a private or internal repository.

Tests verify that:

- A GitHub API release request returning `404` falls back to JIT authentication.
- The request succeeds when retried with the JIT credential.
- Requests without matching static credentials do not trigger JIT fallback.
- Cached JIT credentials are reused without requesting another token.
- Existing GitHub API authentication behavior remains unchanged.

The full test suite passes with:
```
go test ./... -count=1
```

### Checklist

### Rollout and failure behavior

- GitHub API JIT fallback is disabled by default and can be enabled with `PROXY_GITHUB_API_JIT_ACCESS=true`.
- Fallback remains limited to requests authenticated by the proxy.
- GitHub API `404` responses remain potential authentication failures because private repositories can be hidden behind a `404`; when enabled, an ordinary `404` can therefore cause one JIT attempt for that repository.
- Repositories that exhaust all authentication paths are negatively cached so subsequent API requests do not repeat JIT access requests.

### DIFF
--- a/internal/handlers/github_api.go
+++ b/internal/handlers/github_api.go
@@ -52,7 +52,7 @@ func NewGitHubAPIHandler(creds config.Credentials, client ScopeRequester) *GitHu
 		}
 	}
 
-	if len(handler.credentials.data) == 0 {
+	if handler.credentials.isEmpty() && len(handler.jitAccessByHost) == 0 {
 		logrus.Warn("GitHubAPIHandler has no app access tokens")
 	}
 
@@ -85,7 +85,7 @@ func (h *GitHubAPIHandler) HandleRequest(req *http.Request, ctx *goproxy.ProxyCt
 	if !h.isHandledGitHubAPIRequest(req) {
 		return req, nil
 	}
-	if len(h.credentials.data) == 0 {
+	if h.credentials.isEmpty() {
 		return req, nil
 	}
 
--- a/internal/handlers/github_api.go
+++ b/internal/handlers/github_api.go
@@ -19,7 +19,9 @@ import (
 // This allows git credentials for "github.com" to apply to "api.github.com" and
 // will allow git credentials for "<tenant>.ghe.com" to apply to "api.<tenant>.ghe.com" in Proxima.
 type GitHubAPIHandler struct {
-	credentials *gitCredentialsMap
+	credentials     *gitCredentialsMap
+	jitAccessByHost map[string]jitAccessConfig
+	client          ScopeRequester
 }
 
 const ghAPIAddedAuthCtxKey = "gh-api.added-auth"
@@ -27,20 +29,27 @@ const reservedProximaIdentity = "proxima-service-identity"
 
 // NewGitHubAPIHandler returns a new GitHubAPIHandler, extracting the app
 // access token from the array of credentials
-func NewGitHubAPIHandler(creds config.Credentials) *GitHubAPIHandler {
+func NewGitHubAPIHandler(creds config.Credentials, client ScopeRequester) *GitHubAPIHandler {
 	handler := GitHubAPIHandler{
-		credentials: newGitCredentialsMap(),
+		credentials:     newGitCredentialsMap(),
+		jitAccessByHost: map[string]jitAccessConfig{},
+		client:          client,
 	}
 
 	for _, cred := range creds {
-		host := cred.Host()
-		if host == "" {
-			continue
+		switch cred["type"] {
+		case "git_source":
+			host := cred.Host()
+			if host == "" {
+				continue
+			}
+			if host != "github.com" && !strings.HasSuffix(fmt.Sprint(host), ".ghe.com") {
+				continue
+			}
+			handler.credentials.addGitSourceCredentials("api."+host, cred)
+		case "jit_access":
+			handler.addJITAccess(cred)
 		}
-		if cred["type"] != "git_source" || (host != "github.com" && !(strings.HasSuffix(fmt.Sprint(host), ".ghe.com"))) {
-			continue
-		}
-		handler.credentials.addGitSourceCredentials("api."+host, cred)
 	}
 
 	if len(handler.credentials.data) == 0 {
@@ -48,6 +57,27 @@ func NewGitHubAPIHandler(creds config.Credentials) *GitHubAPIHandler {
 	}
 
 	return &handler
+}
+
+func (h *GitHubAPIHandler) addJITAccess(cred config.Credential) {
+	if cred.GetString("credential-type") != "git_source" {
+		return
+	}
+
+	host := strings.ToLower(cred.GetString("host"))
+	if host == "" {
+		return
+	}
+
+	if !strings.HasPrefix(host, "api.") {
+		host = "api." + host
+	}
+
+	h.jitAccessByHost[host] = jitAccessConfig{
+		endpoint: cred.GetString("endpoint"),
+		username: cred.GetString("username"),
+		password: cred.GetString("password"),
+	}
 }
 
 // HandleRequest adds auth to a GitHub API request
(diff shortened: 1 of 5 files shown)


honeyankit:
@thavaahariharangit Helm smoke test is failing.

jurre:
I'm digging through wether this opens up any access to resources/endpoints that we would otherwise expect to be shielded off for users. I _think_ it's probably fine, but I want to make sure that I fully understand how this changes the access that Dependabot gets to these private repositories.

jurre (changes requested):
Can we put all of this behind a feature flag so we can test it in isolation, roll it out to customers slowly and revert it quickly if needed?

internal/handlers/github_api.go
```diff
@@ -121,9 +152,59 @@ func (h *GitHubAPIHandler) HandleResponse(rsp *http.Response, ctx *goproxy.Proxy
 		logging.RequestLogf(ctx, "* re-auth'd request returned %d, ignoring response", newRsp.StatusCode)
 		helpers.DrainAndClose(newRsp)
 	}
+
+	// All known credentials have been tried, try to JIT create access credentials.
+	jitCreds := h.getJITCredentialsForRequest(ctx)
```
jurre, line 169: _Commented by Copilot from a review-session with Jurre._

No negative caching. `GitServerHandler` tracks `reposAlreadyTried` (`git_server.go:426-429`, `464`) so a repo that fails every auth path isn't retried. This path has nothing equivalent, so every request to an inaccessible repo issues a fresh JIT request.

dependabot-core hits several endpoints per dependency (`/releases`, `/tags`, `/contents/...`), so one unresolvable dependency costs multiple JIT round trips, and repeated dependencies multiply it.

`jitRateLimit` is `semaphore.NewWeighted(1)` (`apiclient/client.go:51,68`), which caps concurrency at 1 but doesn't cap the count. With `httpTimeout = 10 * time.Second`, a job resolving a few dozen unreachable repos serialises into minutes of added wall clock.

Can we add a `threadsafe.Map` of failed org/repo pairs, same as `GitServerHandler`?

internal/handlers/github_api.go
```diff
@@ -84,10 +114,11 @@ func (h *GitHubAPIHandler) HandleResponse(rsp *http.Response, ctx *goproxy.Proxy
...
-	if addedAuth, ok := ctxdata.GetBool(ctx, ghAPIAddedAuthCtxKey); !ok || !addedAuth {
+	if !h.isHandledGitHubAPIRequest(ctx.Req) {
 		return rsp
 	}
-	if !h.isHandledGitHubAPIRequest(ctx.Req) {
+	addedAuth, _ := ctxdata.GetBool(ctx, ghAPIAddedAuthCtxKey)
+	if !addedAuth && h.jitAccessByHost[helpers.GetHost(ctx.Req)].endpoint == "" {
 		return rsp
 	}
 	if !isPotentialAuthFailure(rsp.StatusCode) {
```
jurre, line 129: _Commented by Copilot from a review-session with Jurre._

`isPotentialAuthFailure` counts 404 as an auth failure. That's right for private repos, but it fires on a lot of ordinary responses. `GET /repos/{o}/{r}/releases/latest` returns 404 whenever a repo has no published releases, and `/contents/{path}` returns 404 for any missing file. Those are normal results for public repos, and they now trigger JIT requests that can never succeed, since the monolith only grants JIT for internal repos.

Combined with the missing negative cache, ReleaseFinder running against public dependencies becomes the dominant source of JIT traffic. Worth measuring before rollout.

internal/handlers/github_api.go
```diff
@@ -84,10 +114,11 @@ func (h *GitHubAPIHandler) HandleResponse(rsp *http.Response, ctx *goproxy.Proxy
 	if rsp == nil {
 		return rsp
 	}
-	if addedAuth, ok := ctxdata.GetBool(ctx, ghAPIAddedAuthCtxKey); !ok || !addedAuth {
+	if !h.isHandledGitHubAPIRequest(ctx.Req) {
 		return rsp
 	}
-	if !h.isHandledGitHubAPIRequest(ctx.Req) {
+	addedAuth, _ := ctxdata.GetBool(ctx, ghAPIAddedAuthCtxKey)
+	if !addedAuth && h.jitAccessByHost[helpers.GetHost(ctx.Req)].endpoint == "" {
```
jurre: _Commented by Copilot from a review-session with Jurre._

This widens the trigger and I'd like it called out in the PR description.

Before, `HandleResponse` returned early unless the handler had added auth itself. Now, whenever a `jit_access` credential exists for the host, the JIT path runs even for requests the proxy never authenticated (`HandleRequest` returns early when `getCredentialsForRequest` finds nothing).

So any 404 on `/repos/{org}/{repo}` reaches `getJITCredentialsForRequest`, and the org/repo come straight from the request path. Was that intentional, or should the JIT path stay behind `addedAuth`?

internal/handlers/github_api.go
```diff
@@ -19,43 +19,73 @@ import (
...
+	}
+}
+
 // HandleRequest adds auth to a GitHub API request
 func (h *GitHubAPIHandler) HandleRequest(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
 	if !h.isHandledGitHubAPIRequest(req) {
 		return req, nil
 	}
-	if len(h.credentials.data) == 0 {
+	if h.credentials.isEmpty() {
```
jurre: _Commented by Copilot from a review-session with Jurre._

`isEmpty()` checks whether the host map has entries, but `gitCredentialsMap.get()` inserts an empty `hostCredentialMap` on miss (`git_server.go:67-78`). Both `getCredentialsForRequest` and `getJITCredentialsForRequest` call `get()`, so after the first request to a host this returns false even when no tokens were ever added.

Behaviour still works out because `getCredentialsForRequest` returns an empty slice, but `isEmpty()` doesn't mean what the name suggests. Either have it check for actual tokens, or drop it and rely on the `len(creds) == 0` check below.

internal/handlers/github_api.go
```diff
@@ -121,9 +152,59 @@ func (h *GitHubAPIHandler) HandleResponse(rsp *http.Response, ctx *goproxy.Proxy
...
+	if jitConfig.endpoint == "" {
+		return nil
+	}
+
+	org, repo, ok := gitHubAPIExtractOrgAndRepo(ctx.Req.URL.Path)
+	if !ok {
+		return nil
+	}
+
+	logging.RequestLogf(ctx, "* requesting JIT access for github api request")
```
jurre: _Commented by Copilot from a review-session with Jurre._

Nit: the "requesting JIT access" log fires before the `h.client == nil` check, so it logs a request that never happens. Same ordering exists in `git_server.go:503-506`, so fixing both or neither is fine by me.

internal/handlers/github_api_test.go
```diff
@@ -533,3 +533,142 @@ func TestGitHubAPIHandler_TokenFallback_In_Proxima(t *testing.T) {
 		})
 	}
 }
+
+type testGitHubAPIScopeRequester struct {
+	callCount int
+	result    *config.Credential
+}
+
+func (t *testGitHubAPIScopeRequester) RequestJITAccess(ctx *goproxy.ProxyCtx, endpoint string, username string, password string, account string, repo string) (*config.Credential, error) {
```
jurre, line 544: _Commented by Copilot from a review-session with Jurre._

The stub always returns a credential and never an error. A case where `RequestJITAccess` returns `(nil, err)` would cover the retry path and would pin down whatever negative caching we add.

jurre (commented):
_Commented by Copilot from a review-session with Jurre._

The JIT plumbing mirrors `GitServerHandler` closely and the happy-path tests are solid. My concerns are all about what happens when JIT *fails* rather than when it succeeds. `GitServerHandler` guards against retry storms with `reposAlreadyTried`; this handler has no equivalent, and combined with treating 404 as an auth failure that turns routine GitHub API 404s into JIT round trips.

PR #180, opened 2026-08-04
fix CONNECT handling for plain HTTP on port 80

thavaahariharangit:

Fix CONNECT requests carrying plain HTTP traffic on port 80. Previously, goproxy always attempted a TLS handshake, producing first record does not look like a TLS handshake and potentially hanging update jobs.

This upgrades `github.com/elazarl/goproxy` to `v1.8.5`, which detects whether tunneled traffic is HTTP or HTTPS before applying MITM handling.

The dependency upgrade includes upstream https://github.com/elazarl/goproxy/pull/743. No proxy-specific workaround was added; the behavior is covered by an integration-style regression test.

`TestProxyCONNECTPort80CarriesPlainHTTP` reproduces the issue by sending plain HTTP through a CONNECT tunnel. It failed with the previous dependency and passes with `v1.8.5`.

The full test suite passes with go test workspaces..

### Checklist


v-robaiken:
@thavaahariharangit we already have a goproxy bump in https://github.com/dependabot/proxy/pull/175. As far as I can tell the only difference here is the added test, which could go in a follow up if we want it.

If bumping goproxy is the right fix for this issue, could we just use the Dependabot bump PR instead?

jakecoffman (approved):

PR #194, opened 2026-08-10
Fix caching of bodyless HTTP responses

v-thavaahariharangit:

Prevent the proxy cache from turning bodyless HTTP responses into responses with a wrapped body.

When using newer versions of `goproxy`(https://github.com/elazarl/goproxy), wrapping the body of a 304 Not Modified response can cause it to be serialized with chunked body data. This corrupts persistent HTTP connections and causes clients such as Cargo to fail with:

```
[8] Weird server reply (Invalid status line)
```

The change preserves `http.NoBody` for HEAD requests and responses with statuses that do not permit a body, including `1xx`, `204`, and `304`.

This is a defensive fix at the cache boundary. Although `goproxy` should enforce bodyless response semantics during serialization, the cache should not wrap or cache a body that HTTP semantics prohibit.

The response’s original body is still closed so resources are released correctly.

Regression tests verify that bodyless responses:
- Retain `http.NoBody`
- Do not get wrapped by the cache
- Close the original response body
- Do not create a cache entry
The focused cache tests and full Go test suite pass.

The original failure was reproduced using Dependabot job `1404166863`. After applying this change and rerunning the same job with the updated proxy image, the Cargo update completed without the `Invalid status line` error.

using this PR Branch: https://github.com/dependabot/proxy/pull/184
`set -o pipefail && dependabot update -f input.yml --proxy-image=my-proxy-fix -o output.yml 2>&1 | tee tofile.txt`
After fix:
```
updater | 2026/08/10 10:12:47 INFO Finished job processing
updater | 2026/08/10 10:12:47 INFO Results:
updater | +------------------------------------------+
updater | |   Changes to Dependabot Pull Requests    |
updater | +---------+--------------------------------+
updater | | updated | sha2 ( from 0.10.9 to 0.11.0 ) |
updater | +---------+--------------------------------+
  proxy | 2026/08/10 10:12:50 Skipping sending metrics because api endpoint is empty
```

### Checklist


kbukum1 (changes requested):
Before this we need to understand breaking change happened in the goproxy so accordingly we should be able to check the root cause of this error happening

kbukum1:
> Before this we need to understand breaking change happened in the goproxy so accordingly we should be able to check the root cause of this error happening

@thavaahariharangit 
I don't think skipping/bypassing empty responses at the cache boundary is the right solution here — it feels like we're patching a symptom rather than fixing the root cause.

My concern is that the real problem most likely isn't that we're legitimately receiving bodyless/empty responses. It seems more likely that something in the new proxy behavior is causing crates.io to return these responses in the first place — possibly because our implementation is now generating far more requests than before, triggering rate limiting or similar server-side behavior. So instead of defensively wrapping/skipping these responses, we should understand what actually changed in `goproxy` v1.9.0 and adapt our implementation to it properly.

In summary, I think we need to:

- **Audit our implementation against the new version** — compare our current proxy code against the changes in the bump (where the vendored diffs are visible) and confirm our implementation is still fully valid under new version 1.8.5, 1.8.6, 1.9.0.
 
- **Decide the fix based on the root cause** — if refactoring is needed, determine whether the issue is purely on our side (a runtime issue that compiles fine but behaves incorrectly) and refactor accordingly, or whether the library itself also needs a fix.
- **Escalate upstream if the issue is in the library** — if the root cause is in the new `goproxy` code rather than ours, we should open an issue with the maintainer to see if they can address it.

v-thavaahariharangit:
closing this PR, as this implementation is covered here: https://github.com/dependabot/proxy/pull/200

PR #195, opened 2026-08-10
Migrate ECR authentication to AWS SDK for Go v2

JamieMagee:

Move ECR registry authentication from `aws-sdk-go` to `aws-sdk-go-v2` and remove the v1 dependency. The proxy still reads the AWS access key ID and secret from the existing Docker registry username and password fields, so users do not need to change their configuration.

AWS SDK v2 no longer provides `ecriface`, so the handler now defines the single ECR operation it uses. The request context is passed through config loading and `GetAuthorizationToken`. Most of the diff comes from regenerating the vendor tree.

The ECR handler tests cover credential mapping, request context propagation, token reuse, and host matching.

- `go test -race -shuffle=on -count=2 ./...`
- `go build ./...`
- `go vet ./...`
- `go mod tidy -diff`

### Checklist


internal/handlers/docker_registry.go
```diff
@@ -165,7 +169,7 @@ type dockerRegistryCredentials struct {
 	getECRClient getECRClient
 }
 
-func (c *dockerRegistryCredentials) getECRCredentials(ctx *goproxy.ProxyCtx) bool {
+func (c *dockerRegistryCredentials) getECRCredentials(requestCtx context.Context, ctx *goproxy.ProxyCtx) bool {
```
jeffwidman: This takes two contexts? 😕 

Do we need to do any inspection at any point to honor context.cancel on this new context? So that if a request is cancelled, we stop what we're doing?

Or are we relying on other libraries to internally check for cancel, and all we need to do is pass it along?

jeffwidman (dismissed):
Pretty straightforward.

Had one comment about a confusing thing, non-blocking though as it certainly isn't worse than it already is since we don't even plumb the context through currently...

internal/handlers/docker_registry.go
JamieMagee: `requestCtx` is the request's `context.Context`. `proxyCtx` is goproxy's request state and logger, not another cancellation context. We pass `requestCtx` to both `LoadDefaultConfig` and `GetAuthorizationToken`, so the SDK handles cancellation for those calls. There is no other blocking work here that needs a separate `ctx.Err()` check. I renamed the parameters to make that clearer.

internal/handlers/docker_registry.go
```diff
@@ -118,7 +121,7 @@ func (h *DockerRegistryHandler) HandleRequest(req *http.Request, ctx *goproxy.Pr
 			continue
 		}
 
-		if cred.getECRCredentials(ctx) {
+		if cred.getECRCredentials(req.Context(), ctx) {
```
jeffwidman, line 124: Discussed on a call, it's super confusing that we've got this ProxyCtx name floating around using the idiomatic `ctx` which normally _always_ means a go `context.Context` var... 

So we should rename it, I'll throw an agent at it shortly (let's merge this PR first)

jeffwidman (approved):

internal/handlers/docker_registry.go
jeffwidman: * https://github.com/dependabot/proxy/pull/196

PR #197, opened 2026-08-10
Use explicit contexts for network requests

JamieMagee:

Resolve all 266 `noctx` findings. Test requests now use `t.Context()`, proxy tests send explicit requests with `client.Do`, and the connectivity probe uses `DialContext`.

This keeps cancellation available while preserving the probe's 100 ms timeout and cached result.

Most of the diff is mechanical. Handler tests use a small `newTestRequest` helper to avoid repeating the context setup.

The full linter still reports 104 existing `bodyclose`, `errcheck`, `gocritic`, and `gosec` findings. `origin/main` reports the same 104 findings, plus the `noctx` findings fixed here.

- `go test ./...`
- `golangci-lint run --enable-only noctx --max-issues-per-linter=0`

### Checklist

### DIFF
--- a/internal/handlers/test_helpers.go
+++ b/internal/handlers/test_helpers.go
@@ -1,7 +1,9 @@
 package handlers
 
 import (
+	"io"
 	"net/http"
+	"net/http/httptest"
 	"strings"
 	"testing"
 
@@ -10,6 +12,11 @@ import (
 
 	"github.com/dependabot/proxy/internal/config"
 )
+
+func newTestRequest(t *testing.T, method, target string, body io.Reader) *http.Request {
+	t.Helper()
+	return httptest.NewRequestWithContext(t.Context(), method, target, body)
+}
 
 // handleRequestAndClose calls handler.HandleRequest and closes any response body.
 // Most handlers return nil responses, but the linter can't prove that.
--- a/internal/handlers/test_helpers.go
+++ b/internal/handlers/test_helpers.go
@@ -1,9 +1,7 @@
 package handlers
 
 import (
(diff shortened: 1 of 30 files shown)


internal/handlers/test_helpers.go
```diff
@@ -11,6 +13,11 @@ import (
 	"github.com/dependabot/proxy/internal/config"
 )
 
+func newTestRequest(t *testing.T, method, target string, body io.Reader) *http.Request {
+	t.Helper()
+	return httptest.NewRequestWithContext(t.Context(), method, target, body)
```
jeffwidman: nit: I'm actually not a fan of this, I think that using the plain `httptest.NewRequestWithContext(t.Context()...` isn't that much more verbose than `newTestRequest(t...` and make it more clear at the call site what's actually happening...

When I see a custom `newTestRequest`, I wonder "oh, what magic is that concealing?" vs when I see a `httptest.NewRequestWithContext(` I know immediately what's happening

If we were doing more in here such as other setup/teardown/side effects, this would make sense to me but we're not... 

This isn't something I feel strongly about, if you really prefer it, i don't mind approving as-is

jeffwidman: also, in the olden days when we actually coded, this mattered less since I'd zip around inside my editor, but so much of my life these days is either reviewing PR's quickly in the web, or telling an agent what to do, that the friction of looking something up takes more time than it used to... even with the web UI's symbol lookups, it's still something I forget how to do half the time it seems.

jeffwidman (approved):

PR #200, opened 2026-08-12
Upgrade goproxy to v1.9.0

v-thavaahariharangit:

Upgrade `github.com/elazarl/goproxy` from the pinned pseudo-version to `v1.9.0` while preserving Dependabot Proxy's HTTP and HTTPS behavior.

goproxy v1.9.0 writes intercepted HTTPS responses through Go's standard `http.Response.Write` path. This enforces HTTP framing semantics more strictly than the previous manual response writer. The response cache wrapped every response body, including responses that must not carry a body, which could cause invalid transfer framing for HEAD, informational, 204, and 304 responses.

This PR updates and vendors goproxy v1.9.0, adapts the cache to the new response contract, and adds wire-level regression coverage for the upgraded MITM path.

The cache now bypasses and normalizes responses that cannot contain a body. It closes any non-sentinel body, uses `http.NoBody`, and clears transfer-encoding state. A `101 Switching Protocols` response is returned untouched because its body is the upgraded connection stream.

An ordinary response with an unexpected nil body is logged and left uncached. This is distinct from both a transport failure, where goproxy supplies no response, and a valid non-nil zero-byte response body.

HTTP/2 MITM remains disabled. Enabling it would require separate work because request streams may share mutable data stored in `ProxyCtx.UserData`.

Unit coverage verifies HEAD, informational, 204, 304, 101, unexpected nil-body, and valid zero-byte response behavior at the cache boundary.

End-to-end tests exercise fixed-length, chunked, trailer-bearing, bodyless, conditional ETag, cached, and subsequent HTTPS responses through a real CONNECT tunnel. A transport-failure test verifies that an upstream close produces a controlled 500, is logged distinctly from a valid bodyless response, and is not cached.

The following validation passes:

```text
go test ./internal/cache
go test .
go test ./...
go test -race ./...
script/test
```

### Checklist


internal/cache/handlers.go
```diff
@@ -214,6 +214,18 @@ func (d *DB) OnResponse(resp *http.Response, proxyCtx *goproxy.ProxyCtx) *http.R
 		logrus.Warnln("Received nil response")
 		return resp
 	}
+	if resp.StatusCode == http.StatusSwitchingProtocols {
+		return resp
+	}
+	if responseMustNotHaveBody(resp) {
```
kbukum1, line 220: It will be great when we do guard no body response we also log a warning that there is no body response to make it observable.

v-thavaahariharangit:
## goproxy v1.9.0 compatibility assessment

The cache was the only incompatibility found in an active Dependabot Proxy code path. In addition to fixing that incompatibility, we reviewed the other upstream changes that intersect with this repository. Active paths are guarded and covered by focused tests; new opt-in features remain disabled or unused.

| Upstream change | Dependabot Proxy impact | Guard or evidence in this PR |
| --- | --- | --- |
| HTTP/1 MITM response framing and streaming | Active and behaviorally significant | End-to-end HTTPS tests cover fixed-length, chunked, trailer-bearing, HEAD, 204, 304, cached, conditional ETag, and subsequent responses on one proxy connection. |
| Trailer forwarding | Active | The HTTPS MITM test verifies that an upstream response trailer reaches the client. |
| Bodyless responses | Required a cache adaptation | The cache bypasses 1xx, HEAD, 204, and 304 responses, normalizes them to `http.NoBody`, and clears transfer-encoding state. |
| `101 Switching Protocols` | Must retain the upgraded stream | The cache returns `101` unchanged and a focused test verifies that the stream is neither wrapped nor closed. |
| Unexpected ordinary nil body | Invalid response shape at the cache boundary | The cache logs the condition and returns without creating a cache entry; a valid non-nil zero-byte stream remains cacheable. |
| Upstream transport failure | Active | An upstream-close test verifies a controlled `500`, distinct nil-response logging, no cached failure, and successful recovery on the next request. |
| HTTP/2 MITM | New but inactive | `AllowHTTP2` remains at its default `false`. Enabling it is intentionally separate work because concurrent streams would require reviewing mutable `ProxyCtx.UserData`. |
| WebSocket handling refactor | Potentially relevant through upgrades | The cache guard preserves the `101` upgraded stream. The rest of the WebSocket implementation remains owned by goproxy and is not modified here. |
| Per-request dialer and `ConnectionErrHandler` | Opt-in and unused | Existing transport and safe-dialer configuration is retained and passes the complete suite. |
| Header canonicalization and `KeepAcceptEncoding` | Opt-in and unused | Existing defaults are retained. |
| Host-matching changes | No direct dependency | Dependabot handlers use repository-owned host-matching helpers. |
| CA and signer refactor | Existing public integration remains active | Real HTTPS CONNECT tests validate certificate generation and trust using the configured CA. |
| HAR and concurrency-limiter extensions | Unused | Neither extension is imported or enabled by Dependabot Proxy. |
| Removed or renamed APIs | No repository usage | The complete repository builds and tests against v1.9.0 without compatibility shims. |
| Go 1.24 requirement and dependency updates | Compatible | Dependabot Proxy uses Go 1.26; local, race, vendored, and container builds pass. |

Validation completed:

```text
go test ./internal/cache
go test .
go test ./...
go test -race ./...
script/test
```

So the compatibility position is:

- **Adapted:** cache/body-framing behavior.
- **Behaviorally verified:** active HTTP/1 MITM framing, trailers, connection reuse, ETag revalidation, cache behavior, and transport failure.
- **Source-compatible:** existing public goproxy APIs and upstream internal refactors.
- **Intentionally inactive:** HTTP/2 MITM and other opt-in extensions.

v-thavaahariharangit:
I am closing this, as this implementation is being covered here: https://github.com/dependabot/proxy/pull/201

PR #201, opened 2026-08-13
Bump goproxy to v1.9.0 and fix keep-alive framing desync

kbukum1:

Upgrade `github.com/elazarl/goproxy` to **v1.9.0**.

The bump on its own re-introduces a bug we saw with earlier goproxy versions: some jobs fail with errors like Cargo's `[8] Weird server reply (Invalid status line)`. This PR bumps goproxy **and** fixes that failure so the upgrade is safe to ship.

**Root cause (short version):**
- Older goproxy closed the client↔proxy tunnel after every response. v1.9.0 keeps it alive and reuses it for many requests (a good performance change).
- Our cache attaches a body to **every** response, including ones that must have no body (e.g. `304`, `204`, `HEAD`). That produced a few stray bytes the client never reads.
- Before, the connection closed each time, so those bytes were harmless. With the reused connection, they now corrupt the **next** request on the same tunnel → the failures above.

This affects **any** ecosystem that gets these responses, not just Cargo.

- The fix targets the **root cause** (the cache), not the symptom: bodyless responses now stay bodyless through both cache paths. It keys off HTTP status/method only, so it's ecosystem-agnostic.
- We intentionally **keep** goproxy's new keep-alive behavior (the performance win) rather than reverting to always-close.

- New tests reproduce the bad case for `304` / `204` / `HEAD` across both cache paths and assert the response stays correctly framed.
- Full `go vet` + `go test ./...` pass.

### Checklist


internal/cache/handlers.go
```diff
@@ -229,12 +278,27 @@ func (d *DB) OnResponse(resp *http.Response, proxyCtx *goproxy.ProxyCtx) *http.R
 		return resp
 	}
 
+	if bodyless(resp.StatusCode, resp.Request.Method) {
+		// Bodyless responses (1xx/204/205/304/HEAD) have no body to stream to
+		// disk. Leave resp.Body untouched (it is http.NoBody) so goproxy frames
```
jeffwidman: Is there benefit to being defensive here by double-checking that it's `http.NoBody` ?

jeffwidman (dismissed):
Non-blocking observation:
> Our cache attaches a body to every response, including ones that must have no body (e.g. 304, 204, HEAD). That produced a few stray bytes the client never reads.

It sounds like this is a bug already present on `main` regardless of the goproxy bump.

In that case, it might be simpler/safer to have a PR fixing just this, to ensure it works as expected pre-bump. And only then do the separate bump of the proxy (which you could probably do via :dependabot: if you wanted).

 I'm also fine with shipping as-is (I noted you kept the bump change separate from the behavioral change as two distinct commits, thanks!)

It's probably faster to ship both, but if we do revert for some other bug we might trip upon in the proxy, we should probably only revert the proxy bump, not this wrapper fix.

internal/cache/handlers.go
kbukum1: It seems like it may hit. Example `PythonIndexHandler.HandleResponse`  replaces  resp.Body  even for  http.NoBody  on 2xx (so a HEAD/204/205 on a Python index path), and that runs before the cache.

kbukum1:
> Non-blocking observation:
> 
> > Our cache attaches a body to every response, including ones that must have no body (e.g. 304, 204, HEAD). That produced a few stray bytes the client never reads.
> 
> It sounds like this is a bug already present on `main` regardless of the goproxy bump.
> 
> In that case, it might be simpler/safer to have a PR fixing just this, to ensure it works as expected pre-bump. And only then do the separate bump of the proxy (which you could probably do via :dependabot: if you wanted).
> 
> I'm also fine with shipping as-is (I noted you kept the bump change separate from the behavioral change as two distinct commits, thanks!)
> 
> It's probably faster to ship both, but if we do revert for some other bug we might trip upon in the proxy, we should probably only revert the proxy bump, not this wrapper fix.

Actually we don't have a bug if we don't use the bumped one, because it never hits the no-response situation. No response only happens because of request interference on the same connection. If we don't bump, we can't even prove we fixed anything — so I think we should have both together so I can test properly when I deploy, and revert as well.

proxy_test.go
```diff
@@ -46,6 +51,303 @@ func TestProxyHTTPRequest(t *testing.T) {
...
 
+func TestProxyHTTPSMITMFixedLengthResponseFraming(t *testing.T) {
+	upstream := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Length", "5")
+		_, err := io.WriteString(w, "hello")
+		assert.NoError(t, err)
+	}))
+	defer upstream.Close()
+
+	t.Setenv("PROXY_CACHE", "false")
```
jeffwidman, line 62: observation (not something to necessarily fix)

these env vars are global to the process, so when `go test` runs tests, it runs them in parallel for each package (serial within a package, parallel across packages) so it can introduce races.

I don't see a way around that here, and IMO this is just fine/convenient to use the env var... just mentioning it though.

jeffwidman (dismissed):
I am not super familiar with this code/networking protocols that it's touching, but all the explanation here makes sense. 

I think we should try this in production, and see how it behaves. If there are further issues, we can fix forwards as we go, or upstream if needed changes back to `goproxy`.

v-thavaahariharangit (dismissed):

jeffwidman (approved):

PR #204, opened 2026-08-14
Don't cache conditional/bodyless responses; serve conditional requests from cached 200

kbukum1:

Dependabot pnpm jobs were failing with:

```
[ERR_PNPM_CACHE_MISSING_AFTER_304] Metadata cache for @vitest/pretty-format is unreadable after receiving 304 Not Modified
```

The proxy cache was **storing and replaying `304 Not Modified`** (and other bodyless) responses. Replaying a stale `304` tells a client "use your local copy" when, in a fresh job, it may not have that copy — breaking pnpm. There is also no body to reuse, so caching a `304` saves nothing while risking that the cache masks upstream package updates from the updater.

**Never cache `1xx` / `204` / `205` / `304`** (`cacheableStatus`). These carry no reusable body, and a replayed `304` is actively harmful for a dependency updater. They are passed straight through, correctly framed (the bodyless-body normalization from #201 still runs).

The policy is enforced on **both** paths:
- **Writes** (`OnResponse`) — non-cacheable statuses are never stored.
- **Loads** (`New`) — conditional/no-content entries persisted in a `db.yaml` by an older proxy are dropped on load, so they are never replayed after upgrade.

Conditional-request validators (`If-None-Match` / `If-Modified-Since`) are deliberately **kept in the cache key**. An earlier revision collapsed conditional requests onto the unconditional key to serve them from a cached `200`; that was dropped because it could serve a client a representation *older* than the one it already holds, bypassing revalidation and masking upstream updates. Keeping validators in the key means a conditional request is never short-circuited — it either misses and revalidates upstream, or matches its own cached entry.

New and updated tests, all passing (`go test ./...`, `-race`, `go vet` clean):

- `Test_cacheableStatus` — status-code policy table.
- `TestCache_PersistedNonCacheableEntriesAreNotLoaded` — a `db.yaml` containing `304`/`204` entries loads only the cacheable `200`.
- `TestProxyHTTPSConditionalNotModifiedIsNotCached` — every conditional request reaches upstream (no stale `304` replay) over a single, healthy reused MITM tunnel.
- `TestProxyHTTPSConditionalNotModifiedPreservesCachedResponse` — a passthrough `304` does not corrupt the tunnel and leaves the cached unconditional `200` intact.
- Updated `TestCache_BodylessResponses` / `TestCache_BodylessResponseWithWrappedBodyIsRestored` — bodyless bodies are still normalized to `http.NoBody`, but conditional/no-content statuses are no longer cached (only HEAD for a cacheable status is stored, headers-only).

### Checklist

### DIFF
--- a/internal/cache/handlers.go
+++ b/internal/cache/handlers.go
@@ -62,15 +62,6 @@ var ignoreHeaders = map[string]struct{}{
 	"X-Pub-Os":          {},
 	"X-Pub-Reason":      {},
 	"X-Pub-Session-Id":  {},
-	// Conditional-request validators: collapse a conditional request onto the
-	// same key as its unconditional twin so it can be served from an already
-	// cached full 200 body instead of going upstream. We never cache a 304
-	// (see cacheableStatus), so this key only ever holds a full response; a
-	// conditional request therefore either hits that body or passes through
-	// unchanged. Excluding them from the hash never forces a re-download of a
-	// body the client already has.
-	"If-None-Match":     {},
-	"If-Modified-Since": {},
 }
 
 // generates the key used in the DB, includes a hash of the body
@@ -159,6 +150,14 @@ func New(enabled bool, cacheDir string) (*DB, error) {
 		return db, nil
 	}
 	for i := range in {
+		if in[i].Entry != nil && !cacheableStatus(in[i].Entry.Status) {
+			// Skip conditional / no-content entries (1xx/204/205/304) persisted
+			// by an older proxy. cacheableStatus only gates new writes, so
+			// without this filter such entries would still be loaded here and
+			// replayed by OnRequest after upgrade, reintroducing the stale-304
+			// problem this change is meant to fix.
+			continue
+		}
 		db.cacheDB[in[i].Key] = in[i].Entry
 	}
 	// prevent successive runs from overwriting previous cache entries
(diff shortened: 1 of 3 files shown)


kbukum1:
Closing this PR — the reported `ERR_PNPM_CACHE_MISSING_AFTER_304` was already fixed on `main` by #201 (the goproxy v1.9.0 keep-alive **framing** fix). The extra "don't cache conditional/bodyless (`1xx`/`204`/`205`/`304`) responses" change added here is unnecessary for that fix and actively regresses CI.

### Why it fails CI

The smoke jobs download a pre-recorded proxy cache and expect ~100% cache hits (no real registry calls). Several ecosystems — bundler (compact-index), cargo, composer, hex, pip-compile — rely on **conditional requests whose recorded responses are `304 Not Modified`**. This PR dropped those entries on load (`New`) and refused to serve/re-cache them, so those requests missed and hit live registries, returning newer versions than the pinned expectations.

Evidence from this run:
- Failing `e2e (bundler)`: `137/347 calls cached (39%)` → 210 real calls leaked upstream → version drift in the diff (e.g. `json 2.18.1` → `2.21.2`).
- Passing `e2e (npm-pnpm)`: `422/436 calls cached (96%)`.

### Takeaway

In this proxy, **caching and replaying `304` is required and safe** once framing is correct (#201): it's core to the hermetic smoke-test replay and to production cache-hit rates. The theoretical "stale 304 masks an update" concern raised in review conflicts with that replay design and would need a different mechanism (freshness/TTL) rather than dropping 304s.

Leaving the branch `kbukum/no-cache-conditional-responses` in place for reference. If we ever see the 304 symptom again on top of #201, we can reopen and investigate from here.

kbukum1:
I am re-evaluating this option and will reopen if necessary

PR #230, opened 2026-09-02
Add egress control through allowlist to prevent source code exfiltration

AbhishekBhaskar:
Add domain-based egress control to the proxy to help prevent source-code exfiltration during Dependabot update jobs. Even when an attacker abuses a legitimately-configured registry token (e.g. Artifactory/Nexus), traffic to unknown hosts like `evil.com` should be visible and, eventually, blockable — something the earlier Method-header approach could not address.

This PR lands the **observe (log-only)** foundation plus the per-job **dynamic allowlist** needed before enforcement can be turned on: it builds a per-job allowlist and logs any request to a host not on it, without changing traffic behavior by default.

- **Two independent experiment flags** (not a config block): `proxy_egress_observe` (log non-allowlisted requests, still allow) and `proxy_egress_enforce` (drop with **403**, matching the existing `blockMetadataAPIHosts` response). Both default `false` ⇒ **fail-open**, so existing jobs and GHES are unaffected. Using experiments lets us roll this out gradually without a config-schema change.
- **Static base allowlist (union, no `PACKAGE_MANAGER` dependency):** GitHub/Dependabot infrastructure domains plus the **union of every ecosystem's default registry/CDN hosts**, applied to every job. This is deliberate — launchers don't reliably set `PACKAGE_MANAGER` (the action doesn't; the CLI needed patching), and multi-ecosystem jobs need registries beyond their nominal package manager. All entries are trusted public registries, so the containment trade-off is negligible versus the exfil protection (unknown hosts are still blocked).
- **Domains live in an embedded YAML** (`internal/handlers/egress_allowlist_defaults.yaml`, `//go:embed`) parsed at `init()`. The per-ecosystem keyed map is retained there for provenance/documentation; the handler consumes the deduped, sorted union. A `&docker_registries` YAML anchor shares the container-registry set across `docker`, `docker_compose`, `devcontainers`, and `helm`.
- **Dynamic per-job hosts** (derived once from `cfg.Credentials` at construction — no per-request cost):
  - **Configured registries:** the host of every URL-bearing credential field (`host`, `url`, `index-url`, `registry`, `api-host`, `repo`, `endpoint`, `replaces-base`), so a job's private registries (Artifactory/Nexus/private npm, etc.) are never treated as exfiltration.
  - **OIDC token-exchange endpoints:** the identity-provider auth hosts the proxy itself contacts when a credential is OIDC-configured — Azure (`login.microsoftonline.com`), AWS (`sts.amazonaws.com`), GCP (`sts`/`iamcredentials`/`www.googleapis.com`), Cloudsmith (`api-host` or `api.cloudsmith.io`). Detection mirrors the identifying keys in `oidc.CreateOIDCCredential`; the eventual registry target is already covered by the credential-host extraction.
  - **Backend-supplied `domains` escape hatch:** an explicit per-credential `domains` list is merged verbatim (bare hosts or leading-dot suffix patterns) for redirect/CDN targets the proxy can't derive.
- **Boundary-safe matching:** `helpers.HostMatchesDomain` uses a leading-dot suffix check so one entry covers subdomains (`.pkg.dev` → `us-docker.pkg.dev`) while lookalikes like `evilnpmjs.org` are rejected. Exact entries and the suffix form both normalize an absolute DNS name (trailing dot), so `registry.npmjs.org.` matches.
- **Handler placement:** registered in `proxy.go` right after request logging and before all credential-injecting handlers, so drops/logs happen before any auth is added. Works for HTTP and MITM'd HTTPS.
- **No metrics in this PR.** Per-host counting was intentionally dropped (high cardinality / wrong tool for "top hosts"); allowlist tuning uses the observe **logs** (`* egress not allowlisted <host>`) via the logging pipeline / Kusto.

- **Unit:** `HostMatchesDomain` boundary matrix (subdomains, case/trailing-dot normalization, lookalike negatives); handler flag matrix (fail-open, observe-logs-allows, enforce-403, observe+enforce); GitHub-infra always allowed; union covers every ecosystem's defaults regardless of `PACKAGE_MANAGER`; absolute-FQDN exact match; embedded-YAML load/dedup guard; collector separates two hosts into distinct series.
- **Integration** (`proxy_test.go`): end-to-end through MITM — non-allowlisted host returns **403** under `enforce`, and **200 + a `egress not allowlisted <host>` log line** under `observe` (asserted from captured logger output).
- Full suite green: `go build ./...`, `go test ./...`, `gofmt`, and `yamllint` all pass.

### Rollout plan

1. **Observe** (this PR): collect logs of non-allowlisted hosts per ecosystem.
2. **Verify** the dynamic allowlist (configured registries + OIDC + backend `domains`) covers real jobs — confirm from logs that enforce *would not* have broken anything.
3. **Enforce** per-ecosystem by flipping `proxy_egress_enforce` (already wired).

### Checklist


internal/handlers/egress_allowlist_defaults.go
```diff
@@ -0,0 +1,71 @@
...
+//
+// Every registered Dependabot ecosystem has an explicit entry below. An empty
+// entry ({}) means the ecosystem reaches only GitHub/Dependabot infrastructure
+// or plain git (covered by githubInfraDomains and the job's own git
+// credentials), or its registries are configured per job. These are:
+// github_actions, submodules, swift, vcpkg, pre_commit (all git/GitHub-hosted)
+// and helm (chart repositories supplied per job). Keys must match the ecosystem
+// internal names registered in dependabot-core (Dependabot::FileFetchers.register).
+var ecosystemDefaultDomains = map[string][]string{
+	"npm_and_yarn":   {"registry.npmjs.org", "registry.yarnpkg.com"},
```
v-robaiken: can we add jsr.io, registry.bower.io to here too

internal/handlers/egress_allowlist_defaults.go
```diff
@@ -0,0 +1,71 @@
...
+	"bun":            {"registry.npmjs.org", "registry.yarnpkg.com"},
+	"pip":            {"pypi.org", "files.pythonhosted.org"},
+	"uv":             {"pypi.org", "files.pythonhosted.org"},
+	"bundler":        {"rubygems.org", "index.rubygems.org"},
```
v-robaiken: gems.rubygems.org

internal/handlers/egress_allowlist_defaults.go
```diff
@@ -0,0 +1,71 @@
...
+	"maven":          {"repo.maven.apache.org", "repo1.maven.org", "plugins.gradle.org", "dl.google.com"},
+	"gradle":         {"repo.maven.apache.org", "repo1.maven.org", "plugins.gradle.org", "dl.google.com"},
+	"sbt":            {"repo1.maven.org", "repo.maven.apache.org", "repo.scala-sbt.org"},
+	"cargo":          {"crates.io", "static.crates.io", "index.crates.io"},
+	"composer":       {"repo.packagist.org", "packagist.org"},
+	"go_modules":     {"proxy.golang.org", "sum.golang.org", "storage.googleapis.com"},
+	"docker":         {"registry-1.docker.io", "auth.docker.io", "index.docker.io", "production.cloudflare.docker.com", "ghcr.io", "mcr.microsoft.com", "quay.io", "public.ecr.aws", ".gcr.io", ".pkg.dev"},
+	"docker_compose": {"registry-1.docker.io", "auth.docker.io", "index.docker.io", "production.cloudflare.docker.com", "ghcr.io", "mcr.microsoft.com", "quay.io", "public.ecr.aws", ".gcr.io", ".pkg.dev"},
```
v-robaiken: registry.hub.docker.com

internal/handlers/egress_allowlist_defaults.go
```diff
@@ -0,0 +1,71 @@
...
+	"rust_toolchain": {"static.rust-lang.org"},
+	"conda":          {"api.anaconda.org", "anaconda.org", "conda.anaconda.org", "repo.anaconda.com"},
+	"nix":            {"channels.nixos.org"},
+	"devcontainers":  {"registry-1.docker.io", "auth.docker.io", "index.docker.io", "production.cloudflare.docker.com", "ghcr.io", "mcr.microsoft.com", "quay.io", "public.ecr.aws", ".gcr.io", ".pkg.dev"},
+	"github_actions": {}, // GitHub-hosted; covered by githubInfraDomains
+	"submodules":     {}, // plain git; covered by githubInfraDomains + job git credentials
+	"swift":          {}, // SwiftPM git dependencies; covered by githubInfraDomains + git
+	"vcpkg":          {}, // baseline is the microsoft/vcpkg git repo; covered by githubInfraDomains
+	"pre_commit":     {}, // hooks are git repos; covered by githubInfraDomains + git
+	"helm":           {}, // chart repositories supplied per job; OCI charts via githubInfraDomains
```
v-robaiken: We need docker regs here too

internal/handlers/egress_allowlist_defaults.go
```diff
@@ -0,0 +1,71 @@
...
+	"julia":          {"pkg.julialang.org"},
+	"rust_toolchain": {"static.rust-lang.org"},
+	"conda":          {"api.anaconda.org", "anaconda.org", "conda.anaconda.org", "repo.anaconda.com"},
+	"nix":            {"channels.nixos.org"},
+	"devcontainers":  {"registry-1.docker.io", "auth.docker.io", "index.docker.io", "production.cloudflare.docker.com", "ghcr.io", "mcr.microsoft.com", "quay.io", "public.ecr.aws", ".gcr.io", ".pkg.dev"},
+	"github_actions": {}, // GitHub-hosted; covered by githubInfraDomains
+	"submodules":     {}, // plain git; covered by githubInfraDomains + job git credentials
+	"swift":          {}, // SwiftPM git dependencies; covered by githubInfraDomains + git
+	"vcpkg":          {}, // baseline is the microsoft/vcpkg git repo; covered by githubInfraDomains
+	"pre_commit":     {}, // hooks are git repos; covered by githubInfraDomains + git
```
v-robaiken: This is just for pre_commit, but what about additional dependencies?

internal/handlers/egress_allowlist_defaults.go
```diff
@@ -0,0 +1,71 @@
...
+// or plain git (covered by githubInfraDomains and the job's own git
+// credentials), or its registries are configured per job. These are:
+// github_actions, submodules, swift, vcpkg, pre_commit (all git/GitHub-hosted)
+// and helm (chart repositories supplied per job). Keys must match the ecosystem
+// internal names registered in dependabot-core (Dependabot::FileFetchers.register).
+var ecosystemDefaultDomains = map[string][]string{
+	"npm_and_yarn":   {"registry.npmjs.org", "registry.yarnpkg.com"},
+	"bun":            {"registry.npmjs.org", "registry.yarnpkg.com"},
+	"pip":            {"pypi.org", "files.pythonhosted.org"},
+	"uv":             {"pypi.org", "files.pythonhosted.org"},
```
v-robaiken: pypi.python.org, pythonhosted.org

proxy.go
```diff
@@ -73,6 +73,9 @@ func newProxyWithCacheDir(envSettings config.ProxyEnvSettings, cfg *config.Confi
 	proxy.OnRequest().DoFunc(logger.logRequest)
 	proxy.OnResponse().DoFunc(logger.logResponse)
 
+	egressAllowlistHandler := handlers.NewEgressAllowlistHandler(cfg, envSettings, metricsClient)
```
v-robaiken: Is there any reason we don't want to use [metricHandler](https://github.com/dependabot/proxy/pull/230/changes#diff-fd4820e2c415e0057b0e67d294303bdde3b98c2731b6dbd1e2b0f79349744499R94)

internal/handlers/egress_allowlist.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (h *EgressAllowlistHandler) isAllowed(host string) bool {
+	for _, entry := range h.allowed {
+		// A leading dot means "this domain and any subdomain"; otherwise the
+		// entry must match the host exactly.
+		if domain, ok := strings.CutPrefix(entry, "."); ok {
+			if helpers.HostMatchesDomain(host, domain) {
+				return true
+			}
+		} else if helpers.AreHostnamesEqual(host, entry) {
```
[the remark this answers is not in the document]
AbhishekBhaskar: Resolved

internal/handlers/egress_allowlist.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// recordNotAllowlisted emits a counter tagged with the real request host. The
+// package_manager tag is attached by the metrics client's default tags, so the
+// resulting series answers "top hosts per ecosystem" for allowlist tuning.
+func (h *EgressAllowlistHandler) recordNotAllowlisted(host string) {
+	if h.metrics == nil {
+		return
+	}
+	_ = h.metrics.SendMetric("egress_not_allowlisted_count", "increment", 1, map[string]string{"request_host": host})
```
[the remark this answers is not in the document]
AbhishekBhaskar: Fixed

proxy.go
AbhishekBhaskar: The egress handler deliberately takes `metricSender`  directly because it needs the unbucketed real host and a different metric name (`egress_not_allowlisted_count`), vs `metrics.Handler`'s bucketed `http_request_count`

internal/handlers/egress_allowlist.go
```diff
@@ -0,0 +1,105 @@
...
+// infrastructure domains plus the static per-ecosystem defaults for the job's
+// package manager. The observe/enforce toggles are driven by job experiments.
+// The metrics client may be nil, in which case no telemetry is emitted.
+func NewEgressAllowlistHandler(cfg *config.Config, env config.ProxyEnvSettings, metrics metricSender) *EgressAllowlistHandler {
+	allowed := append([]string(nil), githubInfraDomains...)
+	allowed = append(allowed, ecosystemDefaultDomains[env.PackageManager]...)
+
+	return &EgressAllowlistHandler{
+		observe: cfg.Experiments.Enabled(egressObserveExperiment),
+		enforce: cfg.Experiments.Enabled(egressEnforceExperiment),
```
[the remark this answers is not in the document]
AbhishekBhaskar: Updated PR description

v-robaiken (approved):

kbukum1 (approved):

PR #240, opened 2026-09-09
Restore eager NuGet service index discovery

brettfo:
### Short description

Restore eager NuGet service-index discovery while retaining the useful credential matching and deduplication behavior introduced after the original implementation.

Paged v2 feeds and v3 service indexes can identify authenticated endpoints that differ from the configured repository URL. The proxy must discover those endpoints before the updater starts, otherwise requests can be made before the proxy knows which credentials to inject.

PR #203 changed NuGet discovery from eager initialization to deferred response interception. That caused update-job failures when authenticated resource URLs were needed before the proxy had observed the service-index response. This PR restores eager initialization and improves its performance, matching, determinism, and redirect handling.

The proxy still completes NuGet discovery before opening port 1080. Discovery requests now run through a bounded pool of eight workers, while results are merged in configured credential order.

### Behavior before PR #203 compared with this PR

| Area | Before PR #203 | Current behavior |
|---|---|---|
| Initialization | Service indexes were fetched eagerly and serially. | Service indexes are fetched eagerly with up to eight concurrent requests. Port 1080 is opened only after all discovery finishes. |
| Discovery client | Used a private client with a 10-second timeout and the default transport. | Uses the proxy's shared 10-second client and restricted transport, including the safe dialer and TLS settings. |
| Unusable credentials | Credentials without a token or password were stored and queried without authentication. | Unusable static credentials are ignored and do not trigger discovery. |
| Duplicate sources | Every configured source was queried, including equivalent duplicates. | Canonically equivalent sources are queried once, with first-configured precedence. HTTP and HTTPS sources remain distinct. |
| Credential precedence | Discovery was interleaved with configuration processing, so an earlier discovered credential could shadow a later explicit credential. | All explicit credentials are registered before discovered credentials, so explicit configuration wins. Duplicate discovered resources use deterministic configured-source order. |
| Path matching | Used decoded string-prefix matching; `/feed` could match `/feedback`, and escaped separators could be conflated with literal separators. The first matching credential won. | Uses canonical path-segment matching, preserves escaped separators, rejects ambiguous dot segments, and selects the most-specific matching URL credential. |
| Host credentials | A host credential could shadow a URL credential based on configuration order. | URL-scoped credentials take precedence; host credentials are fallback matches. |
| Proxy-only credentials | Had no special NuGet behavior and acted like ordinary credentials. | Are host-only fallback credentials used only for HTTPS on the default port when no explicit/OIDC credential or usable request authorization applies. |
| Response handling | Parsed any response below HTTP 400 and could panic on short or empty unrecognized bodies. | Parses only HTTP 200 responses and safely handles empty, short, invalid, and non-success responses. |
| Service-index redirects | Relied on the standard client's redirect-header behavior and did not register redirected index URLs. | Explicitly retains `Authorization` or `X-Api-Key` across redirects, including redirects to another host, and registers every followed redirect URL for later authenticated requests. |
| Discovered resources | Registered service-index resource URLs with the source credential, including resources on other hosts. | Retains that behavior while canonicalizing and deduplicating registrations. |
| OIDC matching | Used decoded path-prefix matching. | Uses canonical, most-specific path matching and participates in concurrent discovery, deduplication, and redirect registration. |

Authentication formatting is unchanged: username/password values use Basic authentication, ordinary tokens use Bearer authentication, and Azure DevOps tokens are treated as Basic-auth passwords. Discovery still supports v2 `xml:base` and non-template v3 resources.

- Authentication is intentionally retained when a service-index redirect changes hosts, matching the treatment of cross-host resource URLs returned by NuGet indexes.
- Discovery concurrency is capped at eight workers. Workers collect results independently; shared credential and OIDC state is updated afterward in deterministic configuration order.
- Canonical static URL deduplication is scheme-agnostic for compatibility, while source discovery keeps HTTP and HTTPS distinct.

- All NuGet resource and redirected service-index URLs are registered before the proxy accepts connections.
- Slow feeds do not serialize discovery of every other configured feed.
- The most-specific applicable credential is selected safely and deterministically.
- Explicit credentials take precedence over discovered credentials.
- Cross-host redirects and cross-host resources retain authentication.
- Empty, malformed, and non-200 discovery responses do not add credentials or panic.
- `go test ./...` passes.

### Checklist

### DIFF
--- a/internal/handlers/nuget_feed.go
+++ b/internal/handlers/nuget_feed.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/elazarl/goproxy"
 
@@ -47,12 +48,23 @@ type nugetFeedCredentials struct {
 	proxyOnly bool
 }
 
+const nugetDiscoveryConcurrency = 8
+
+type nugetDiscoveryJob struct {
+	serviceIndexURL     string
+	staticCredential    nugetFeedCredentials
+	staticCredentialSet int
+	oidcCredential      *oidc.OIDCCredential
+}
+
 // NewNugetFeedHandler returns a new NugetFeedHandler.
 func NewNugetFeedHandler(creds config.Credentials, client *http.Client) *NugetFeedHandler {
 	handler := NugetFeedHandler{
 		credentials:  []nugetFeedCredentials{},
 		oidcRegistry: oidc.NewOIDCRegistry(client),
 	}
+	staticCredentialSets := make([][]nugetFeedCredentials, 0)
+	discoveryJobs := make([]nugetDiscoveryJob, 0)
 
 	for _, cred := range creds {
 		if cred["type"] != "nuget_feed" {
@@ -75,53 +87,12 @@ func NewNugetFeedHandler(creds config.Credentials, client *http.Client) *NugetFe
 		} else {
 			oidcCredential, _, ok := handler.oidcRegistry.Register(cred, []string{"url"}, "nuget feed")
 			if ok {
-				// Discover additional resource URLs from the nuget feed index.
 				if url != "" {
-					func() {
-						req, err := http.NewRequestWithContext(context.Background(), "GET", url, nil)
-						if err != nil {
-							logging.RequestLogf(nil, "error creating http request (%s): %v", url, err)
-							return
-						}
-
-						if req.URL.Scheme != "https" {
-							logging.RequestLogf(nil, "refusing to discover nuget feed over non-https URL %s", url)
-							return
-						}
-
-						if !handler.oidcRegistry.TryAuth(req, nil) {
-							return
-						}
-
-						rawRsp, err := client.Do(req)
-						if err != nil {
-							logging.RequestLogf(nil, "error retrieving http response (%s): %v", url, err)
-							return
-						}
-						defer rawRsp.Body.Close()
-
-						body, err := io.ReadAll(rawRsp.Body)
-						if err != nil {
-							logging.RequestLogf(nil, "error reading http response body (%s): %v", url, err)
-							return
-						}
-
-						switch rawRsp.StatusCode {
-						case 401, 403:
-							logging.RequestLogf(nil, "unauthorized for nuget feed %s", url)
-							return
-						}
-
-						if rawRsp.StatusCode >= 400 {
-							logging.RequestLogf(nil, "unexpected http response %d for nuget feed %s", rawRsp.StatusCode, url)
-							return
-						}
-
-						urlsToAuthenticate := extraUrlsFromSourceResponse(body, url)
-						for _, discoveredURL := range urlsToAuthenticate {
-							handler.oidcRegistry.RegisterURL(discoveredURL, oidcCredential, "nuget resource")
-						}
-					}()
+					discoveryJobs = append(discoveryJobs, nugetDiscoveryJob{
+						serviceIndexURL:     url,
+						staticCredentialSet: -1,
+						oidcCredential:      oidcCredential,
+					})
 				}
 				continue
 			}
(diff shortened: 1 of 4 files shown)


internal/handlers/nuget_feed.go
```diff
@@ -116,26 +110,133 @@ func NewNugetFeedHandler(creds config.Credentials, client *http.Client) *NugetFe
...
+
 func extraUrlsFromSourceResponse(body []byte, url string) []string {
 	var urls []string
 	bodyString := strings.TrimSpace(string(body))
-	if bodyString == "" {
-		logging.RequestLogf(nil, "empty API response from NuGet feed %s", url)
-		return nil
-	}
 	bodyReader := bytes.NewReader(body)
 	switch {
```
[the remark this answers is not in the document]
brettfo: Fixed in 4f37ea4. Empty and whitespace-only bodies now return safely, and short unknown bodies use a precision-limited log format instead of slicing.

internal/handlers/nuget_feed.go
```diff
@@ -116,26 +110,133 @@ func NewNugetFeedHandler(creds config.Credentials, client *http.Client) *NugetFe
...
+
+	switch rawRsp.StatusCode {
+	case http.StatusUnauthorized, http.StatusForbidden:
+		logging.RequestLogf(nil, "unauthorized for nuget feed %s", job.serviceIndexURL)
+		return nil
+	}
+	if rawRsp.StatusCode >= http.StatusBadRequest {
+		logging.RequestLogf(nil, "unexpected http response %d for nuget feed %s", rawRsp.StatusCode, job.serviceIndexURL)
+		return nil
+	}
```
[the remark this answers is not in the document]
brettfo: Fixed in 4f37ea4. Eager discovery now accepts only HTTP 200 responses; redirects and bodyless 204/205 responses are logged and skipped.

internal/handlers/nuget_feed_test.go
```diff
@@ -655,216 +363,117 @@ func TestNugetFeedHandlerProxyOnlyCredentials(t *testing.T) {
...
-func (r *readErrorThenData) Read(p []byte) (int, error) {
-	if r.first != nil {
-		n := copy(p, r.first)
-		r.first = nil
-		return n, assert.AnError
+func nugetDiscoveryResponse(resourceURL string) *http.Response {
+	return &http.Response{
+		StatusCode: http.StatusOK,
+		Body: io.NopCloser(strings.NewReader(fmt.Sprintf(
+			`{"version":"3.0.0","resources":[{"@id":%q,"@type":"PackageBaseAddress/3.0.0"}]}`,
```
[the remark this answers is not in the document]
brettfo: Added regression coverage in 4f37ea4 for empty, whitespace-only, and short unknown response bodies, plus non-OK discovery responses.

jeffwidman:
this is an agent review comment guided by jeff

1. **Cross-origin redirect authentication is a new policy change, not part of restoring eager discovery.** Current `main` authenticates service-index redirects only while they remain same-origin; its regression test requires cross-origin redirects and later hops on that foreign origin to remain unauthenticated, while still registering resources learned from the final index. The pre-#203 eager implementation also used the standard Go client behavior, which strips `Authorization` on redirects to unrelated hosts. PR #240 preserved this behavior until `1410ce1`, which now copies `Authorization` and `X-Api-Key` to every redirect and registers every redirect URL for future authentication. This broadens credential exposure beyond both baselines and also sends credentials across HTTPS-to-HTTP redirects, bypassing the normal HTTPS-only OIDC gate. PR #206 discussed duplicate resource-route claims and did not change the redirect policy on `main`. Please restore the pre-`1410ce1` same-origin guard and test: continue following cross-origin redirects without credentials and parse their final index, but authenticate only same-origin redirect hops. The PR description should likewise stop claiming cross-host redirect authentication as part of this revert.

2. **Scheme-less configured feed URLs no longer get eager discovery.** NuGet matching accepts URLs such as `nuget.example.com/v3/index.json`, and current `main` can learn resources when that configured index is requested over HTTP or HTTPS. PR #240 queues the raw value and passes it to `http.NewRequestWithContext`, which fails with `missing protocol scheme`; the configured route can still authenticate, but delegated resource and redirect routes are never discovered. Please normalize scheme-less sources to an explicit discovery URL while retaining their request-matching behavior, or explicitly preserve the current HTTP/HTTPS discovery semantics another way, and add regression coverage.

brettfo:
Noting for future reference.

1. Cross-origin redirect authentication.  This is added by design.  There's nothing in the NuGet spec that says the service index must return same-origin hosts for the other endpoints and in fact it's common to use a CDN or similar here, especially Azure DevOps Artifacts feeds.  Currently the ADO feeds don't require auth on the `30x` redirects but there's nothing to say that can't happen in the future.
2. Not adding a scheme (`http` or `https`) to registry values is by design.  The only kinds of values allowed by NuGet are either `http`, `https`, or directory paths and since dependabot doesn't mount any user-defined directories, this means only `http` and `https` are allowed and everything else is a configuration error.  This is to force users to explicitly set `https` when appropriate so we don't make an incorrect assumption.

jeffwidman:
this is an agent review comment guided by jeff

Do not retain authentication across an HTTPS-to-HTTP redirect. The redirect callback currently copies `Authorization` and `X-Api-Key` to every destination. An HTTPS service index can therefore cause static credentials or an exchanged OIDC token to be sent over plaintext, bypassing the handler’s normal HTTPS-only OIDC protection. Static credentials are also registered for the HTTP redirect URL and may be sent there on later requests. Please strip authentication for HTTPS-to-HTTP redirects, keep it stripped for the remainder of that redirect chain, and add regression coverage for both static and OIDC credentials.

internal/handlers/nuget_feed.go
```diff
@@ -116,19 +116,176 @@ func NewNugetFeedHandler(creds config.Credentials, client *http.Client) *NugetFe
...
+		responseURL = redirectReq.URL
+		redirectAuthAllowed = redirectAuthAllowed && sameOrigin(via[len(via)-1].URL, redirectReq.URL)
+		if !redirectAuthAllowed {
+			redirectReq.Header.Del("Authorization")
+			redirectReq.Header.Del("X-Api-Key")
+			return nil
+		}
+		result.authenticatedRedirectURLs = append(result.authenticatedRedirectURLs, redirectReq.URL.String())
+		return nil
+	}
```
JamieMagee, line 260: Can we reject non-HTTPS redirects for OIDC discovery? IMO removing auth headers isn't enough here.

internal/handlers/nuget_feed.go
```diff
@@ -116,19 +116,176 @@ func NewNugetFeedHandler(creds config.Credentials, client *http.Client) *NugetFe
...
+
+func discoverNugetFeedURLsForJob(
+	job nugetDiscoveryJob,
+	client *http.Client,
+	oidcRegistry *oidc.OIDCRegistry,
+) nugetDiscoveryResult {
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, job.serviceIndexURL, nil)
+	if err != nil {
+		logging.RequestLogf(nil, "error creating http request (%s): %v", job.serviceIndexURL, err)
+		return nugetDiscoveryResult{}
```
JamieMagee, line 222: Can you make sure to resolve schema-less URls e.g. `feed.example.com/v3/index.json`, to absolute URLs before `http.NewRequestWithContext`? Otherwise discovery fails and later successful index requests won't register credentials for the package endpoints.

internal/handlers/nuget_feed.go
```diff
@@ -116,19 +116,176 @@ func NewNugetFeedHandler(creds config.Credentials, client *http.Client) *NugetFe
...
+	if job.oidcCredential != nil {
+		if req.URL.Scheme != "https" {
+			logging.RequestLogf(nil, "refusing to discover nuget feed over non-https URL %s", job.serviceIndexURL)
+			return nugetDiscoveryResult{}
+		}
+		if !oidcRegistry.TryAuthCredential(req, nil, job.oidcCredential) {
+			return nugetDiscoveryResult{}
+		}
+	} else {
+		authenticateNugetRequest(req, job.staticCredential, nil)
```
JamieMagee, line 234: This bypasses `normaliseHost`, so the case-sensitive Azure check selects Bearer instead of Basic auth for a colonless PAT on `PKGS.DEV.AZURE.COM`.

JamieMagee (approved):

internal/handlers/nuget_feed.go
brettfo: Are schema-less URIs common here?  Technically they're not allowed in `NuGet.Config` so I'd prefer to reject them unless we have a reason to accept them.

brettfo:
In an effort to unblock some scenarios, this will be merged as-is and I'll immediately start work on all remaining comments here.

PR #241, opened 2026-09-11
Generate proxy image attestations

JamieMagee:

Publish GitHub build provenance for proxy images so consumers can verify their source. Build and push both image tags in one step, then attest the returned digest in GitHub and GHCR.

Image tags are published before attestation. If attestation fails, the workflow skips Git tag creation and leaves the image tags published. This replaces the shell publisher with pinned Docker actions and `actions/attest`. Local build/test behavior, platform support and tag formats stay unchanged.

YAML lint and shell checks passed. A local registry build published both image tags with the same digest. Live OIDC issuance still needs an authorized main-branch run. The complete test suite was not run.

### Checklist

- [ ] I have thoroughly tested my code changes to ensure they work as expected, including adding additional tests for new functionality.


README.md
```diff
@@ -10,6 +10,33 @@ To build and run the proxy, you need to have the following installed:
...
+  --signer-workflow dependabot/proxy/.github/workflows/ghcr.yml \
+  --source-ref refs/heads/main \
+  --source-digest "${SOURCE_SHA}"
+```
+
+The publication summary includes the digest, source commit and attestation link.
+Add `--bundle-from-oci` to retrieve the attestation from GHCR instead of the GitHub API.
+
+These attestations cover newly built container images. The workflow does not backfill historical images.
+Runtime verification in the CLI and Action, and attestations for the native CodeQL archives, are separate work.
```
jeffwidman: This entire paragraph seems like it will go stale quickly so not worth including.

script/cibuild-publish-ghcr
```diff
@@ -1,32 +0,0 @@
-#!/bin/bash
-
-set -e
-cd "$(dirname "$0")/.."
-ROOT="$(pwd)"
-source "${ROOT}/script/_common"
-
-function main() {
-  docker_build
-  push_to_ghcr
```
jeffwidman, line 10: It took me a moment to realize that getting rid of this script does _not_ prevent us from building this image locally for dev/testing purposes.

This only gets rid of publishing to GHCR manually, which is probably okay for us... there may be scenarios down the road where we want to manually publish an image to GHCR, for example from a specific branch. 

From what I can see, we couldn't do that today through the GitHub Action, but that wouldn't be hard to add down the road if needed. I'm fine with delaying adding it until we actually need it.

jeffwidman (dismissed):

README.md
JamieMagee: Removed.

script/cibuild-publish-ghcr
JamieMagee: If we start adding attestations, I cannot see a scenario where we want to manually publish. It would be a huge red flag if some images are published without attestations.

PR #252, opened 2026-09-15
Refactor egress allowlist experiment names to dash-case

AbhishekBhaskar:
Fix a dash/underscore key mismatch between the proxy and the API for the egress-allowlist experiment flags.

The proxy read the observe/enforce experiments as `proxy_egress_observe` / `proxy_egress_enforce` (underscores). But the API serializes job experiments through the JSON:API adapter, whose default key transform is **dash** — so in the job-details payload the keys actually arrive as `proxy-egress-observe` / `proxy-egress-enforce`, like every other forwarded experiment. With the underscore keys, `Experiments.Enabled(...)` never matched, so the egress allowlist would silently stay in fail-open mode no matter how the experiment was configured.

This renames the constants to dash-case so the flags actually activate when the experiment is enabled.

- Pure key-string rename — no behavioral/logic change beyond making the flags match what the API sends. The constants are the single source of truth; the handler already keys off them.
- Updated the two proxy-level integration tests (`TestProxyEgressAllowlistEnforceBlocks` / `TestProxyEgressAllowlistObserveAllows`) that hard-coded the old underscore keys — otherwise they'd have silently stopped exercising enforce/observe.
- Added a regression test that pins the contract: it asserts the constants equal the literal dash keys, that `Enabled("proxy-egress-observe")` is true on a dash-keyed experiments map, that the underscore form does **not** match, and that a handler built from the dash-keyed payload enters observe mode.

- New `TestEgressAllowlist_DashCasedExperimentKeys` locks in the dash-key behavior.
- Full suite + `gofmt` / `go vet` pass locally (`go test ./...`).
- End-to-end: with the experiment enabled, the proxy now logs (observe) / blocks (enforce) non-allowlisted hosts instead of staying fail-open.

### Checklist


v-thavaahariharangit (approved):
LGTM

PR #259, opened 2026-09-22
preserve Host header exactly unless it differs from the request URL

### LINKED ISSUE #258: normaliseHost breaks SigV4 pre-signed URL redirects
**Affected versions:** `v2.0.20260918154032` and later
**Regression introduced by:** #253 (merged 2026-09-18T15:39:20Z)

## Summary

Since #253, the proxy overwrites the outgoing `Host` header with `req.URL.Host`, which for
MITM'd CONNECT tunnels carries an explicit `:443`. Amazon S3 pre-signed URLs sign `host` as a
signed header, so the canonical request the upstream computes no longer matches the one the
signer used, and S3 rejects the download with `403 SignatureDoesNotMatch`.

This breaks any package download that is served via a 302 redirect to signed cloud storage —
Artifactory with `downloadRedirect` enabled.

## Root cause

`proxy.go`, changed in #253:

```go
 req.URL.Host = strings.ToLower(req.URL.Host)
-req.Host = strings.ToLower(req.Host)
+// Credential checks and upstream virtual-host routing must agree with the dial destination.
+req.Host = req.URL.Host
```

Previously `req.Host` was the client's own `Host` header, lowercased — for an HTTPS request that
is the bare hostname, because clients omit the default port. Now it is taken from `req.URL.Host`.
For a MITM'd CONNECT tunnel, goproxy reconstructs the absolute URL from the CONNECT target
(`host:443`), so `req.URL.Host` retains the explicit port.

Result: the proxy sends `Host: <bucket>.s3.amazonaws.com:443` where the signer used
`host:<bucket>.s3.amazonaws.com`. Different canonical request, different signature.

The intent of #253 (stop a spoofed `Host` header from overriding the dial destination — see
`TestDependabotAPIHandler_SpoofedHost`) is sound. The problem is only that the normalized value
retains a default port that the client would never have sent.

Requests to the registry itself are unaffected, because HTTP Basic auth does not sign the `Host`
header. Only the signed-URL hop breaks. In our logs the same job shows:

```
[010] 200 https://<tenant>.jfrog.io:443/artifactory/api/npm/npm-all/npm/10.9.4
[012] 302 https://<tenant>.jfrog.io:443/artifactory/api/npm/npm-all/npm/-/npm-10.9.4.tgz
[014] 403 https://<bucket>.s3.amazonaws.com:443/<path>/filestore/61/61e8...
[014] Remote response:
      <Error><Code>SignatureDoesNotMatch</Code>
      <Message>The request signature we calculated does not match the signature you
      provided. Check your key and signing method.</Message>
      <StringToSign>AWS4-HMAC-SHA256
      20260922T073452Z
      20260922/eu-central-1/s3/aws4_request
      &lt;redacted&gt;</StringToSign>
```

## Expected behaviour

The `Host` header the proxy sends upstream should be byte-identical to what the client would have
sent, so that host-signed requests validate. Normalizing case, and refusing a `Host` that
disagrees with the dial destination, should not introduce a default port.

jakecoffman:

- fixes https://github.com/dependabot/proxy/issues/258
- follow-on to https://github.com/dependabot/proxy/pull/253

The previous fix overwrote the Host header immediately to prevent future problems where the Host header and URL didn't match. Unfortunately the Host header is used in Amazon S3 pre-signed URLs.

So to fix that, this PR only overwrites the Host header if the Hostnames differ between them.  

If any are true, the Host header is overwritten:
- `if err != nil` - the Host with the scheme attached is unparsable
- `hostURL.Host != req.Host` - The parsed Host and Host header differ, meaning the Host had a path or some other shenanigans
- `!helpers.AreHostnamesEqual(hostURL.Hostname(), req.URL.Hostname())` - IDNA hostname comparison fails (case insensitive)

I was able to reproduce this with Dependabot CLI using jenkins-ci.org as a registry, so I already know it works.

### Checklist


proxy.go
```diff
@@ -174,8 +176,13 @@ func handleForbidden(rsp *http.Response, proxyCtx *goproxy.ProxyCtx) *http.Respo
...
 	req.URL.Host = strings.ToLower(req.URL.Host)
-	// Credential checks and upstream virtual-host routing must agree with the dial destination.
-	req.Host = req.URL.Host
+	hostURL, err := url.Parse(req.URL.Scheme + "://" + req.Host)
+	// Preserve signed Host headers when their hostname matches the dial destination,
+	// otherwise ensure the Host header matches the dial destination.
+	if err != nil || hostURL.Host != req.Host ||
+		!helpers.AreHostnamesEqual(hostURL.Hostname(), req.URL.Hostname()) {
+		req.Host = req.URL.Host
+	}
```
jurre, line 186: I think this fails the following scenario?

```
{
    name:     "matching hostname with mismatched port",
    url:      "https://EXAMPLE.com:443/path",
    host:     "example.com:8443",
    want:  "example.com:443",
},
```

jakecoffman: I left port out intentionally since it doesn't matter for Host spoofing, but I'll go ahead and add it in since it doesn't hurt anything.

jurre: @jakecoffman I think we might want to something like this:

```
func normaliseHost(req *http.Request, proxyCtx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
	req.URL.Host = strings.ToLower(req.URL.Host)

	hostURL, err := url.Parse(req.URL.Scheme + "://" + req.Host)
	if err != nil ||
		hostURL.Host != req.Host ||
		!helpers.AreHostnamesEqual(hostURL.Hostname(), req.URL.Hostname()) ||
		effectivePort(hostURL) != effectivePort(req.URL) {
		req.Host = req.URL.Host
	}

	return req, nil
}

func effectivePort(u *url.URL) string {
	if port := u.Port(); port != "" {
		return port
	}

	switch strings.ToLower(u.Scheme) {
	case "https":
		return "443"
	case "http":
		return "80"
	default:
		return ""
	}
}
```

That way, we ensure the supplied port matches or is the default for the scheme and I think that should cover all the scenarios?

jakecoffman: fixed in 5c905ec601efe98fec6775c1b00d94ca8a58cf72

proxy_test.go
```diff
@@ -69,29 +70,81 @@ func TestNormaliseHost(t *testing.T) {
...
 		{name: "spoofed host", url: "https://EXAMPLE.com/path", host: "api.dependabot.com", want: "example.com"},
-		{name: "default port", url: "https://EXAMPLE.com:443/path", host: "example.com", want: "example.com:443"},
-		{name: "custom port", url: "https://EXAMPLE.com:8443/path", host: "example.com", want: "example.com:8443"},
+		{name: "implicit HTTPS port", url: "https://EXAMPLE.com:443/path", host: "example.com", want: "example.com"},
+		{name: "explicit HTTPS port", url: "https://EXAMPLE.com/path", host: "example.com:443", want: "example.com:443"},
+		{name: "matching HTTPS port", url: "https://EXAMPLE.com:443/path", host: "example.com:443", want: "example.com:443"},
+		{name: "custom port", url: "https://EXAMPLE.com:8443/path", host: "example.com", want: "example.com"},
+		{name: "matching custom port", url: "https://EXAMPLE.com:8443/path", host: "EXAMPLE.com:8443", want: "EXAMPLE.com:8443"},
+		{name: "mismatched port", url: "https://EXAMPLE.com:443/path", host: "example.com:8443", want: "example.com:8443"},
+		{name: "different scheme default port", url: "http://EXAMPLE.com/path", host: "example.com:443", want: "example.com:443"},
```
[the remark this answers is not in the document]
jakecoffman: fixed in 5c905ec601efe98fec6775c1b00d94ca8a58cf72

proxy.go
```diff
@@ -174,8 +176,13 @@ func handleForbidden(rsp *http.Response, proxyCtx *goproxy.ProxyCtx) *http.Respo
...
 	return req, nil
```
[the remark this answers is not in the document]
jakecoffman: fixed in 5c905ec601efe98fec6775c1b00d94ca8a58cf72

jurre (approved):

PR #263, opened 2026-09-23
Add egress allowlist block flag to distinguish observe-only from enforce-mode

v-abhishekbhaskar:
The `egress_host` metric is currently the only egress signal forwarded to Splunk — the request logs are not. Today it carries `allowlisted=false` for every host that misses the allowlist, but that tag can't distinguish a host that was **actually blocked** (enforce mode returned a 403) from one that was merely **observed and allowed through** (observe-only mode). That makes it impossible to tell, from Splunk alone, whether enforcing the allowlist would have broken real traffic.

This adds a `blocked` tag to the metric, set to `true` only when a host is not allowlisted **and** enforce mode dropped the request. Both cases still emit `allowlisted=false`, so existing discovery queries are unaffected while the new tag adds the enforce-impact dimension.

- `blocked` is deliberately narrow: `!allowed && enforce`. An observe-only miss is `allowlisted=false, blocked=false`; an enforce drop is `allowlisted=false, blocked=true`; an allowlisted host is always `blocked=false`.
- The observation is still recorded before the 403 short-circuits the request chain, so enforce-blocked hosts are captured.

New/updated unit tests assert the tag across all modes:

- `TestEgressAllowlist_RecordsObservedHosts` — allowlisted `blocked=false`; observed miss `blocked=false`.
- `TestEgressAllowlist_RecordsEnforceBlockedHosts` — enforce drop `blocked=true`.
- `TestEgressAllowlist_BlockedTagDistinguishesObserveFromEnforce` — same host yields `blocked=false` under observe vs `blocked=true` under enforce.

Once deployed, a Splunk query filtering `blocked=true` will surface hosts an enforce rollout would actually break.

### Checklist


v-robaiken (dismissed):

internal/handlers/egress_allowlist.go
v-robaiken, line 1: I think we should rename blocked to enforced. It's more explicit about the intention than blocked is.

v-abhishekbhaskar: Renamed to `block_enforced`

v-robaiken (approved):

